PR #1, opened 2016-03-03
Configuration file parsing

cored:
Add config.NewFromString to parse configuration files

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,19 @@
+package config
+
+import "github.com/go-yaml/yaml"
+
+type Config struct {
+	Name string `yaml:"name"`
+}
+
+func NewFromString(s string) (*Config, error) {
+	c := new(Config)
+
+	err := yaml.Unmarshal([]byte(s), &c)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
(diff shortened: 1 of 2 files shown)


bobbytables:
I don't think `config` should be its own package. I think it should be a part of a bigger thing in Broadway.

bobbytables:
I think a `broadway` glue package makes sense.

cored:
Could you elaborate what do you mean by a bigger thing?

bobbytables:
A `config` package is a pretty early abstraction considering this config isn't even used by anything. I think writing something that requires configuration and pushing this code into whatever _that_ package is makes more sense.

config/config.go
```diff
@@ -0,0 +1,19 @@
...
+type Config struct {
+	Name string `yaml:"name"`
+}
+
+func NewFromString(s string) (*Config, error) {
+	c := new(Config)
+
+	err := yaml.Unmarshal([]byte(s), &c)
+
+	if err != nil {
```
bobbytables: Typically, you'd do this in one line since `Unmarshal` modifies the pointer. So something like this:

```go
if err := yaml.Unmarshal([]byte(s), &c); err != nil {
  return nil, err
}
```

bobbytables:
@cored - Is this still a thing? Closeable?

PR #3, opened 2016-03-09
Docs

macat:


README.md
```diff
@@ -1,2 +1,106 @@
...
+    "version": "dc231ba",
+    "assets_version": "dc231ba",
+    "owner": "bill"
+  }
+}
+```
+
+Response:
+```
+Status: 201 Created
```
bobbytables: This should probably be a 202.

macat: Why 202?

macat: I think it should be 201 if it's a new object and 200 if it's an update.

bobbytables: because a 202 is a deferred create. The effect of this request is not immediately noticeable right?

README.md
```diff
@@ -1,2 +1,106 @@
 # broadway
 Broadway 
+
+Broadway helps developers deploy their projects with defining their deployment
+workflows in Broadway playbooks. It runs as a service interacting with
+Kubernetes and allowing users to interact with it through various interfaces
+(Slack, CLI, Web).
```
bobbytables: :+1: :fire:

cored: Broadway allows that but through the API, not specified to any client

README.md
macat: The resource (instance) will be created. The user need fire off a deploy call to deploy it.

README.md
macat: What?

README.md
bobbytables: Ah I see.

README.md
bobbytables: Saying I like it!

bobbytables:
:+1:

README.md
macat: Rafael had a comment here, that's why I was asking.

poopoothegorilla:
+1

PR #4, opened 2016-03-09
Playbook parsing: Separate disk read and parse from main function, add first tests

therealplato:
- Move functionality out of main into `ReadPlaybookFromDisk` and `ParsePlaybook`
- Rename `ValidateManifests` to `ManifestsPresent` (as it's just checking if the file exists on disk for now)
- Test `ReadPlaybookFromDisk` with a mock playbook yaml file
- Test `ParsePlaybook`
- Test `ManifestsPresent`


main.go
```diff
@@ -41,31 +41,41 @@ func (t Task) ValidateManifests() error {
...
 			return err
 		}
 	}
 	return nil
 }
 
+func ParsePlaybook(playbook []byte) (Playbook, error) {
+	var p Playbook
+	err := yaml.Unmarshal(playbook, &p)
+	return p, err
```
therealplato: is it correct to return `p` here instead of `&p`?

macat: yes, that'll copy the value instead of returning a pointer.

main.go
```diff
@@ -41,31 +41,41 @@ func (t Task) ValidateManifests() error {
...
+}
+
+func ReadPlaybookFromDisk(fd string) ([]byte, error) {
+	file, err := ioutil.ReadFile(fd)
+	return file, err
```
bobbytables: probably could just do `return ioutil.ReadFile(fd)` instead of assignment then return.

main.go
```diff
@@ -41,31 +41,41 @@ func (t Task) ValidateManifests() error {
...
+}
+
 func main() {
 	args := os.Args
-	yamlFile := args[1:][0]
+	yamlFileDescriptor := args[1:][0]
```
bobbytables: I think we should move to codegangsta for our CLI tool. It makes parsing arguments, help messages, etc, super easy to do.

poopoothegorilla:
should we update these to match the other testing style in @macat code using testify?

macat:
+1

poopoothegorilla:
+1

PR #5, opened 2016-03-09
Adding store and instance + dependencies

macat:


instance/instance.go
```diff
@@ -0,0 +1,39 @@
+package instance
+
+import "encoding/json"
+
+type InstanceStatus string
+
+const (
+	InstanceStatusNew       InstanceStatus = ""
```
poopoothegorilla: how does this work? or is this a miss type?

poopoothegorilla: oh nvm

poopoothegorilla:
+1

PR #7, opened 2016-03-10
Create Instance API

macat:


bobbytables:
@macat - Can you rebase and squash commits so we don't have things like "save" in the history?

macat:
@bobbytables  
done

bobbytables:
:shipit:

PR #8, opened 2016-03-10
Validate playbook and tasks

therealplato:
This PR adds validation logic for a playbook yaml file. The following conditions are required for `playbook.Validate()` to return without err:
* that the Playbook has a populated Name field
* that the Playbook has more than zero items in Tasks array
* that each Task has a populated Name field
* that each Task has either a PodManifest string, or more than zero items in Manifests array
* that each PodManifest and Manifests is a filename that exists on disk

The yaml format has changed: pod_manifest was an array of strings, and is now a string.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -105,24 +105,63 @@ func TestParsePlaybookIncomplete(t *testing.T) {
 }
 
 func TestValidatePlaybook(t *testing.T) {
-	ParsedPlaybook1, _ := ParsePlaybook(PlaybookBytes)
-	if err := ParsedPlaybook1.Validate(); err != nil {
-		t.Error(err)
+	ValidTask := Task{
+		Name:      "task",
+		Manifests: []string{ManifestFilename},
+	}
+	InvalidTask1 := Task{
+		Manifests: []string{ManifestFilename},
+	}
+	InvalidTask2 := Task{
+		Name: "task",
 	}
 
-	ParsedPlaybook2, _ := ParsePlaybook(IncompletePlaybookBytes)
-	if err := ParsedPlaybook2.Validate(); err == nil {
-		t.Errorf("Validation of incomplete playbook succeeded, expected error")
-	}
-
-	ParsedPlaybook3 := ParsedPlaybook1
-	ParsedPlaybook3.Tasks = []Task{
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -57,12 +56,17 @@ func TestMain(m *testing.M) {
 	f, _ := os.Create(PlaybookFilename)
 	f.Write(PlaybookBytes)
 	f.Close()
+
+	f, _ = os.Create(ManifestFilename)
```
bobbytables: Don't dump this err, we need to check every error.

main_test.go
```diff
@@ -57,12 +56,17 @@ func TestMain(m *testing.M) {
 	f, _ := os.Create(PlaybookFilename)
```
bobbytables: Err check

main_test.go
```diff
@@ -101,24 +105,86 @@ func TestParsePlaybookIncomplete(t *testing.T) {
...
+	for _, testcase := range testcases {
+		playbook := testcase.playbook
+		err := playbook.Validate()
+		if testcase.errExpected && err != nil {
+			continue
+		} else if testcase.errExpected && err == nil {
+			t.Errorf("Scenario %s: Succeeded, expected failure", testcase.scenario)
+		} else if err != nil {
+			t.Errorf("Scenario %s: Failed with %s, expected success", testcase.scenario, err)
+		}
```
bobbytables: Because in this test table loop, there's no checking of the error that we're actually expecting, it's mostly implicit.

main_test.go
```diff
@@ -101,7 +112,73 @@ func TestParsePlaybookIncomplete(t *testing.T) {
...
+			},
+			"Task requires at least one manifest or a pod manifest",
+		},
+	}
+
+	for _, testcase := range testcases {
+		playbook := testcase.playbook
+		err := playbook.Validate()
+		if testcase.expectedErr == "" && err == nil { // expected success, got success
+		} else if testcase.expectedErr != err.Error() { // expected failure, got wrong failure
```
bobbytables: Couldn't this be reduced to a single if?

bobbytables:
:+1

gxb5443:
:+1:

PR #11, opened 2016-03-11
Load Playbooks on broadway boot

therealplato:
This PR will iterate through a hardcoded `./playbooks` directory (to be overridden by a CLI flag in a future PR.)

The broadway executable will attempt to parse each file in this directory into a Playbook struct.
It will then check the list of tasks' manifests exist in `.manifests/`, and will assume that `.yml` should be added to the manifest names.
A top level `id` field in a playbook yaml file will now be required.


playbook/playbook.go
```diff
@@ -0,0 +1,125 @@
+package playbook
+
+import (
+	"errors"
```
macat: import lines should follow an order:
- standard library
- third party

`go-import` can help you doing this.

therealplato: my vim-go plugin reorders them! i assumed it was following a standard pattern. i'll look for how to configure it

playbook/playbook_test.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+func TestMain(m *testing.M) {
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Switch cwd to project root before creating fixtures:
+	newCwd := filepath.Join(cwd, "..")
+	os.Chdir(newCwd)
```
macat: Not sure why this is necessary.

therealplato: `fixtures.SetupTestFixtures` is being invoked from two different working directories, for `main_test.go` and `playbook/playbook_test.go`, this was my solution to pointing to the same absolute path from each

macat:
+1

PR #14, opened 2016-03-15
Implement API calls: GET instance, GET instances

therealplato:


therealplato:
It's not obvious to me how to set up a mock instance to test the success case. I would prefer to avoid actually calling `POST /instances` if it's possible to create one without a gin context

instance/base.go
```diff
@@ -45,6 +45,24 @@ func Get(playbookId, id string) (Instance, error) {
...
+		var attrs InstanceAttributes
+		err := json.Unmarshal([]byte(v), &attrs)
+		if err != nil {
+			return nil, err
+		}
+		i := &baseInstance{
+			attributes: &attrs,
+			store:      s,
+		}
+		instances = append(instances, i)
```
therealplato: @macat I'm still unclear on why I can return this slice of pointers (`&baseInstance`) and have them fulfil the return type of `[]Instance` interface

macat: `baseInstance` implements the `Instance` interface.

instance/base.go
```diff
@@ -61,6 +79,14 @@ func (instance *baseInstance) Attributes() *InstanceAttributes {
 	return instance.attributes
 }
 
+func (instance *baseInstance) MarshalJSON() ([]byte, error) {
+	o, err := instance.Attributes().JSON()
+	if err != nil {
+		return []byte{}, err
```
bobbytables: this can be `nil`.

server/server_test.go
```diff
@@ -88,3 +89,106 @@ func TestCreateInstanceWithInvalidAttribtes(t *testing.T) {
...
+	}
+	if len(okResponse) != 2 {
+		t.Errorf("Expected 2 instances matching playbook testPlaybookFull, actual %v\n", len(okResponse))
+	}
+}
+
+func TestGetInstancesWithEmptyPlaybook(t *testing.T) {
+	w := httptest.NewRecorder()
+
+	req, _ := http.NewRequest("GET", "/instances/testPlaybookEmpty", nil)
```
bobbytables: don't dump the error.

server/server_test.go
```diff
@@ -88,3 +89,106 @@ func TestCreateInstanceWithInvalidAttribtes(t *testing.T) {
...
+
+	mem := store.New()
+
+	server := New(mem).Handler()
+	server.ServeHTTP(w, req)
```
bobbytables: This probably returns an err no?

therealplato: @bobbytables Apparently not https://github.com/gin-gonic/gin/blob/master/gin.go#L259

macat:
+1

instance/base.go
```diff
@@ -45,35 +47,73 @@ func Get(playbookId, id string) (Instance, error) {
 	return instance, nil
 }
 
-func (instance *baseInstance) ID() string {
-	return instance.Attributes().Id
+// List looks up all Instances stored under a given playbookID
+func List(s store.Store, playbookID string) ([]Instance, error) {
```
cored: @therealplato Why is this in `base.go`?

therealplato: Get was already here, since List does something similar I put it in the same place

macat:
Rename base.go to default.go

instance/instance.go
```diff
@@ -2,25 +2,33 @@ package instance
...
+// Status represents the lifecycle state of one instance
+type Status string
 
 const (
-	InstanceStatusNew       InstanceStatus = ""
-	InstanceStatusDeploying                = "deploying"
-	InstanceStatusDeployed                 = "deployed"
-	InstanceStatusDeleting                 = "deleting"
-	InstanceStatusError                    = "error"
+	// StatusNew represents a newly created instance
```
cored: @therealplato Could you elaborate on the intention of this? I really don't get it. Like if you are creating a new instance why is the status bound to that particular struct?

server/server.go
```diff
@@ -34,19 +44,23 @@ func New(s store.Store) *Server {
 func (s *Server) setupHandlers() {
 	s.engine = gin.Default()
 	s.engine.POST("/instances", s.createInstance)
```
therealplato: I think this should be POST /instance

instance/instance.go
bobbytables: How is this bound to a struct? Do you mean it being bound to the `Status` type defined above?

server/server.go
macat: I think `/instances` is fine.

instance/instance.go
cored: @bobbytables yes; sorry that's what I meant

bobbytables: So this might be an improper abstraction for right now, but! It does make sense if we ever want to retrieve other values from the Status type. For example, if wanted to make:

```go
func (s Status) IsAvailable() bool {}
```

macat: If you want to make sure `Attributes.Status` can only receive values defined as `Status`. If `Attributes.Status` was a string for example, you could set any value.

macat:
+1

macat:
Great work!

cored:
:+1:

PR #15, opened 2016-03-15
Add command line for creating instance

cored:


Godeps/Godeps.json
```diff
@@ -1,11 +1,13 @@
 {
 	"ImportPath": "github.com/namely/broadway",
 	"GoVersion": "go1.6",
-	"Packages": [
-		"./..."
-	],
 	"Deps": [
 		{
+			"ImportPath": "github.com/codegangsta/cli",
```
macat: you are not using this.

commands/root.go
```diff
@@ -0,0 +1,69 @@
...
+	RootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+}
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
+	if cfgFile != "" { // enable ability to specify config file via flag
+		viper.SetConfigFile(cfgFile)
+	}
+
+	viper.SetConfigName(".broadway") // name of config file (without extension)
```
therealplato: do we need to swap the order here? In case of any conflicts, a config passed in on command line should override the default config location

cmd/create.go
```diff
@@ -0,0 +1,53 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("create called")
+	},
+}
+
+var instanceCmd = &cobra.Command{
+	Use:   "instance",
+	Short: "broadway create instance PLAYBOOK_ID ID [FLAGS]",
+	Run: func(cmd *cobra.Command, args []string) {
+		ia := instance.Attributes{PlaybookID: args[0], ID: args[1], Vars: map[string]string{"foo": "bar"}}
```
bobbytables: `if len(args) < 2` should probably be in here somewhere right? If we try accessing `args[1]` with nothing in the slice this program will panic hard.

cmd/version.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/kardianos/osext"
+	"github.com/spf13/cobra"
+)
+
+// versionCmd represents the version command
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print the version number of Broadway",
+	Long:  `All software has versions. This is Broadway's`,
```
bobbytables: We should use double quotes here to be consistent.

cmd/version.go
```diff
@@ -0,0 +1,57 @@
...
+		return ""
+	}
+	fi, err := os.Lstat(filepath.Join(dir, filepath.Base(fname)))
+	if err != nil {
+		fmt.Println(err)
+		return ""
+	}
+	t := fi.ModTime()
+	return fmt.Sprintf("VERSION: %s\n", t.Format(time.RFC3339))
+}
```
bobbytables: Can you explain what this function is doing?

cored: That's something that I will remove. It is just checking for the binary
creation date.

bobbytables:
Lets avoid code rot please. Can we get this up-to-snuff and merge?

cored:
This is the wrong implementation closing it.

PR #16, opened 2016-03-18
instance.GetStatus implementation, API

therealplato:
This PR will implement and test `instance.GetStatus(store, playbookID, instanceID)`

It will add an HTTP API handler for `status/:playbookID/:instanceID`


macat:
+1

PR #17, opened 2016-03-21
Run golint, go vet, errcheck on CI. Switch Alpine to Debian.

therealplato:
* I have not yet found any way to exclude the `vendor` folder when running against `./...`, or a way to specify multiple packages in one command, thus the many commands
* `go vet` and `errcheck` return 1 when they find problems; `golint` returns 0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 FROM golang:1.6.0-alpine
 RUN apk add --update git && rm -rf /var/cache/apk/*
 
-RUN go get github.com/tools/godep
+RUN go get github.com/tools/godep github.com/kisielk/errcheck github.com/golang/lint/golint
 
 RUN mkdir -p /go/src/github.com/namely/broadway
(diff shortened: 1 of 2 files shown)


cored:
it would be nice to have this as an external service something similar to hound, but I'm always down to increase the quality of the code, so :+1:

macat:
This is open source, so we can hook this up with hound. Let me do it.

macat:
Added hound!

macat:
+1

rickyxiv:
Why did we switch from Alpine to Debian?

therealplato:
@rickyxiv: @macat can give you more info, I believe he was running into kernel panics caused by alpine

macat:
@rickyxiv I had a problem when I compiled kubernetes code. Tried different go versions, different kubernetes versions, nothing helped. Then switched to Debian and it started running again. It was segfaulting after launching the application without any reason or warning. Asked kub devs on slack and they said it must be some kind of go bug.

PR #18, opened 2016-03-21
Infrastructure to handle Slack commands

therealplato:
This PR implements parsing and validation functionality needed to control a Broadway server from a Slack command.

It expects a `POST /command` with a x-www-form-urlencoded payload as specified by https://api.slack.com/slash-commands.

It sends 501 Unauthorized if the token in the POST request does not match an environment variable.

Additionally it sends 200 OK if Slack sends `GET /command?ssl_check=`.

It returns a help message given `/broadway help` but for all other commands it returns a dummy response (i.e. slack commands won't yet call methods on Broadway services.)


server/server.go
```diff
@@ -132,3 +148,58 @@ func (s *Server) getStatus(c *gin.Context) {
...
+	var form SlackCommand
+	if err := c.BindWith(&form, binding.Form); err != nil {
+		c.JSON(http.StatusInternalServerError, InternalError)
+		return
+	}
+	if form.Token != s.slackToken {
+		c.JSON(http.StatusUnauthorized, UnauthorizedError)
+		return
+	}
+	if form.Command != "/broadway" {
```
bobbytables: I don't recommend this check because if we're receiving this request, odds are they have set it up explicitly. An organization may want to use `/bway` in their slack configuration for example.

server/server_test.go
```diff
@@ -14,6 +16,39 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var testToken = "BroadwayTestToken"
+
+func TestServerNew(t *testing.T) {
+	_, exists := os.LookupEnv(slackTokenENV)
+	if exists {
+		t.Fatalf("Found existing $%s. Skipping tests to avoid changing it...", slackTokenENV)
+	}
```
bobbytables: Can you explain whats going on here?

therealplato: If the token environment variable has been set outside the test environment, I don't want to overwrite it, in case the user set their real token then ran the tests

bobbytables: Seems weird, can we inject in the token somehow on the store object?

therealplato: yes, at the cost of never testing "the server loads the token from ENV on startup"

bobbytables: So I think if the env is already set, that it's not terrible to just check what it was set for instead of failing outright, right?

bobbytables: Or even swap it for this test and change it back afterwards using `defer`

bobbytables: Wait.. now that I'm thinking about it, we can just set it. It won't change the persons environment variable for ever, only for the test run, which should be fine.

therealplato: if you like, we can store it and restore the original value during teardown, it would get lost if there's a panic but it's probably not worth worrying about that

PR #19, opened 2016-03-22
Arch refactor

cored:
This PR introduce some concepts to facilitate the introduction of new features to Broadway: 

It introduce the notion of services so every use case for any particular feature will be map directly into one of those. Also add the notion of a repository to separate persistence logic from the the entities. I want some feedback on this ideas.


domain/instance.go
```diff
@@ -0,0 +1,57 @@
+package domain
```
bobbytables: Honestly, I'm not siked on the `domain` package name, I understand it's intention, but if I was just traversing this package looking for something and I saw a "domain" folder I wouldn't think about it as a DDD choice. If we're putting our domain entities in this package, I'd rather opt for `package broadway`.

cored: @bobbytables Agree with that, good sir. Will rename it.

domain/instance.go
```diff
@@ -0,0 +1,57 @@
...
+	encoded, err := json.Marshal(i)
+	if err != nil {
+		return "", err
+	}
+	return string(encoded), nil
+}
+
+// Path for an instance
+func (i *Instance) Path() string {
+	return "/broadway/instances/" + i.PlaybookID + "/" + i.ID
```
therealplato: this might be an antipattern for uncle bob's clean architecture - isn't this a dependency from the inner circle (the abstract Instance object) out to an outer circle (the etcd storage implementation)? ditto for `Save()`

domain/instance.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Instance repository interface
+type InstanceRepository interface {
```
bobbytables: I think this should go in its own file called `repository.go`.

domain/instance.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"encoding/json"
+
+	"github.com/namely/broadway/store"
+)
+
+// Instance entity
+type Instance struct {
+	PlaybookID string            `json:"playbook_id" binding:"required"`
+	ID         string            `json:"id"`
```
therealplato: should this have `binding:"required"`? we're not yet testing any create/get/save instance without id

domain/instance.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Instance repository interface
+type InstanceRepository interface {
+	Save(instance Instance) error
+}
+
+// Instance repository to handle persistence logic
+type InstanceRepo struct {
+	store store.Store
+}
```
bobbytables: You should do a type check here for making sure `InstanceRepo` conforms to the `InstanceRepository` interface. Like so:

```go
var _ InstanceRepository = &InstanceRepo{}
```

This gets evaluated at compile time and ensures that `InstanceRepo` is implementing everything correctly.

domain/instance.go
bobbytables: Typically entities in DDD don't define logic like validation.

domain/instance.go
cored: @therealplato I'm totally aware of this and I agree but until we keep separating the inner components in smaller pieces we are not going to have the right abstraction. Until then we can keep adding this type of code smells for now.

bobbytables: I can see your point, but I think that might be a little too hardcore. This doesn't reveal direct behavior, you could think of this as an identifier.

Perhaps the method name is that is making this hit the 4th wall. Calling this method `Key()` might make more sense.

domain/instance.go
cored: @bobbytables I had the same idea but I just leave it in here for the sake of convenience.

domain/instance.go
cored: @therealplato Sure

server/server.go
```diff
@@ -59,21 +61,23 @@ func (s *Server) Run(addr ...string) error {
 }
 
 func (s *Server) createInstance(c *gin.Context) {
-	var ia instance.Attributes
-	var err = c.BindJSON(&ia)
+	var i domain.Instance
+	var err = c.BindJSON(&i)
 	if err != nil {
```
bobbytables: If a method only returns an error, you should be doing an inline err check. So in this case specifically:

```go
if err := c.BindJSON(&i); err != nil {
  ...
}
```

domain/instance.go
cored: @bobbytables Oh right; you mention that trick from our last pairing session. Will do.

domain/instance.go
cored: @bobbytables Since I'm just refactoring and trying not to change a lot of the current code. But yes, I agree with that.

domain/instance.go
cored: @bobbytables :-) well, you already know where I'm going ;-). I'm having this same issue in my head right now this type of validation should be part of the service and I'll move it but I need another piece for that puzzle.

server/server.go
cored: @bobbytables Gotcha

domain/instance.go
cored: @bobbytables Thinking about it; I won't introduce this line. I feel it will introduce confusion because it doesn't show intent. Right now my code is type checked because I'm using interfaces every where so if I pass an struct without implementing properly the interface it will fail the test.

bobbytables: It is intention revealing to a Gopher. :smile:

cored: I think that the reason that you need that is because you are defining a concrete struct as an argument in a function instead of just using the interface.

server/server.go
```diff
@@ -59,21 +61,22 @@ func (s *Server) Run(addr ...string) error {
...
+	if err := c.BindJSON(&i); err != nil {
 		c.JSON(422, InvalidError("Missing: "+err.Error()))
 		return
 	}
 
-	i := instance.New(s.store, &ia)
-	err = i.Save()
+	repo := broadway.NewInstanceRepo(store.New())
+	service := services.NewInstanceService(repo)
+	err := service.Create(i)
```
therealplato: If we're making a new repo, new instance service, and calling create for each new instance - why not skip the instance service and call `services.CreateInstance(store.New(), i)`?

cored: @therealplato The layer that handles the persistence is the repository one the one for the store is just a representation of where we are going to save the data. The services only orchestrate the calls to the internals of the domain model.

server/server.go
```diff
@@ -62,36 +65,38 @@ func (s *Server) Run(addr ...string) error {
...
-	}
+	service := services.NewInstanceService(s.store)
+	i, err := service.Show(c.Param("playbookID"), c.Param("instanceID"))
 
-	c.JSON(http.StatusOK, i.Attributes())
+	if err != nil {
+		switch err.(type) {
+		case broadway.InstanceNotFoundError:
+			c.JSON(http.StatusNotFound, NotFoundError)
+			return
```
therealplato: @bobbytables these returns are necessary, gin will concatenate multiple sequential `c.JSON()` calls if you don't break control flow after you send the response

macat:
+1

therealplato:
+1

PR #20, opened 2016-03-24
Deployment

macat:


deployment/deployment.go
```diff
@@ -0,0 +1,78 @@
...
+
+func init() {
+
+	scheme := runtime.NewScheme()
+	v1.AddToScheme(scheme)
+	factory := serializer.NewCodecFactory(scheme)
+	deserializer = factory.UniversalDeserializer()
+
+	kcfg := &restclient.Config{
+		Host:     "http://localhost:8080",
```
cored: @macat Shouldn't this be an environment variable?

deployment/deployment.go
```diff
@@ -0,0 +1,78 @@
...
+type Step interface {
+	Deploy() error
+}
+
+// DefaultStep implements a deployment step
+type DefaultStep struct {
+	object runtime.Object
+}
+
+var _ Step = &DefaultStep{}
```
therealplato: Why do you create and dump the DefaultStep?

deployment/deployment.go
macat: OOo yes, it will be. This is super WIP.

deployment/deployment.go
macat: It's just declaring that the defaultstep implements the Step interface.

deployment/deployment.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return s, nil
+}
+
+// Deploy executes the deployment of a step
+func (s *DefaultStep) Deploy() error {
+	oGVK := s.object.GetObjectKind().GroupVersionKind()
+
+	if oGVK.Kind == "ReplicationController" {
+		rc, err := client.ReplicationControllers("default").Create(object.(*v1.ReplicationController))
```
therealplato: is "default" a kubernetes namespace?

macat: It'll be a parameter in the future.

macat: Yes.

deployment/deployment.go
therealplato: Doesn't that just... happen? Since you declare `func (s *DefaultStep) Deploy() error {}`?

macat: Yes, but this way it's more explicit.

cored: @therealplato He is trying to get the compiler to help him. @macat The thing is that I don't think we need that trick if you are doing TDD the compiler will tell you that you are expecting a certain type using an interface and at that moment you will be force to implement those methods.

deployment/deployment.go
```diff
@@ -0,0 +1,74 @@
...
+type Deployment struct {
+	Playbook  playbook.Playbook
+	Variables map[string]string
+	Manifests map[string]*manifest.Manifest
+}
+
+// Deploy executes the deployment
+func (d *Deployment) Deploy() error {
+	for _, task := range d.Playbook.Tasks {
+		if len(task.Manifests) > 0 {
```
bobbytables: Why bother with the length check here?

macat: Because that'll identify what kind of task it is.

bobbytables: We should probably add a method to the `Task` object that can identify that right?

macat: yes, that's a good idea

bobbytables: I'd propose doing a switch statement in here and calling an unexported function on the Deployment instance for whatever needs to happen.

cored: What is the relationship between a task and a manifest I really can't see
it.

PR #23, opened 2016-03-25
Deployment 2

macat:


deployment/deployment.go
```diff
@@ -53,22 +63,44 @@ type Deployment struct {
...
+		err := taskstep.step.Deploy()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// collectSteps collects all the steps for the deployment from the playbook
```
bobbytables: As an fyi, for unexported functions you actually don't NEED to document them. I love when you do though!

cored: What does the steps prefix means?

deployment/deployment_test.go
```diff
@@ -16,41 +16,79 @@ func init() {
...
-		Name: "Test deployment",
-		Meta: playbook.Meta{},
-		Vars: []string{"test"},
-		Tasks: []playbook.Task{
-			{
-				Name: "First step",
-				Manifests: []string{
-					"test",
+	cases := []struct {
+		Tasks    []playbook.Task
```
bobbytables: Can we add a name to our test tables and log them inside of the test table loop please? That way when things break we know exactly which test table entry failed.

deployment/deployment.go
```diff
@@ -53,22 +73,54 @@ type Deployment struct {
...
+		if task.PodManifest != "" {
+			m := d.Manifests[task.PodManifest]
+			rendered := m.Execute(d.Variables)
+			object, err := deserialize(rendered)
+			if err != nil {
+				return steps, err
+			}
+			step := NewPodmanifestStep(object)
+			steps = append(steps, TaskStep{&task, step})
+		} else {
```
therealplato: caution, by my understanding we may see podManifest and Manifests in the same task, can anyone confirm/deny? if so, the else block will skip the Manifests

services/deployment.go
```diff
@@ -0,0 +1,40 @@
...
+func (d *DeploymentService) Deploy(p playbook.Playbook, vars map[string]string) error {
+	return d.deployer.Deploy(p, vars)
+}
+
+// DefaultDeployer implements the Deployer interface
+type DefaultDeployer struct{}
+
+// Deploy finds the playbook's manifests and executes them with Kubernetes.
+func (dd *DefaultDeployer) Deploy(p playbook.Playbook, vars map[string]string) error {
+	MS := NewManifestService()
```
bobbytables: uppercase? What is this? :stuck_out_tongue:

deployment/deployment.go
```diff
@@ -42,6 +55,13 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
+
+	namespace = os.Getenv("KUBERNETES_NAMESPACE")
+}
+
+// Deployer declares something that can Deploy Deployments
+type Deployer interface {
+	Deploy(playbook.Playbook, map[string]string) error
```
bobbytables: What is `map[string]string` here? Vars?

deployment/kubernetes_test.go
```diff
@@ -0,0 +1,111 @@
...
+			Name: "Step with 1 podmanifest file",
+			Tasks: []playbook.Task{
+				{
+					Name:        "First step",
+					PodManifest: "test",
+				},
+			},
+			Expected: 2,
+		},
+	}
```
therealplato: we should add test case for playbook with multiple tasks

services/deployment.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Deploy deploys a playbook
+func (d *DeploymentService) Deploy(instance *broadway.Instance) error {
+
+	log.Println("---")
+	log.Printf("%+v", d.playbooks)
+	log.Println("---")
+	playbook := d.playbooks[instance.PlaybookID]
+
```
therealplato: add err if playbook was missing

services/deployment.go
```diff
@@ -0,0 +1,61 @@
...
+	err := deployer.Deploy()
+	if err != nil {
+		log.Println(err)
+		instance.Status = broadway.StatusError
+	} else {
+		instance.Status = broadway.StatusDeployed
+	}
+	d.repo.Save(instance)
+
+	return nil
```
therealplato: maybe `return err` instead?

services/deployment.go
```diff
@@ -0,0 +1,61 @@
...
+		return errors.New("Instance is being deleted already.")
+	}
+
+	instance.Status = broadway.StatusDeploying
+	d.repo.Save(instance)
+
+	err := deployer.Deploy()
+	if err != nil {
+		log.Println(err)
+		instance.Status = broadway.StatusError
```
macat: return error here.

services/deployment_test.go
```diff
@@ -0,0 +1 @@
+package services
```
macat: add tests

services/manifest.go
```diff
@@ -0,0 +1,98 @@
...
+// implementations
+type ManifestService struct {
+	rootFolder string
+	extension  string
+}
+
+// NewManifestService instantiates a ManifestService with a default rootFolder
+func NewManifestService() *ManifestService {
+	return &ManifestService{
+		rootFolder: "./manifests",
```
therealplato: use env var over default location

therealplato:
+1

PR #24, opened 2016-03-28
Server tests refactoring

cored:


server/server_test.go
```diff
@@ -103,173 +82,57 @@ func TestCreateInstanceWithInvalidAttributes(t *testing.T) {
...
-
-	log.Println(w.Body.String())
-	var okResponse []instance.Attributes
-
-	err = json.Unmarshal(w.Body.Bytes(), &okResponse)
-	if err != nil {
-		t.Error(err)
-		return
-	}
-	if len(okResponse) != 2 {
```
therealplato: We should keep these assertions `assert.Nil(t, err)` and `assert.Len(okResponse, 2)`

server/server_test.go
```diff
@@ -103,173 +82,57 @@ func TestCreateInstanceWithInvalidAttributes(t *testing.T) {
...
 
 func TestGetInstancesWithEmptyPlaybook(t *testing.T) {
-	w := httptest.NewRecorder()
-
-	req, err := http.NewRequest("GET", "/instances/testPlaybookEmpty", nil)
-	if err != nil {
-		t.Error(err)
-		return
-	}
+	req, w := testutils.GetRequest(t, "/instances/testPlaybookFull")
```
therealplato: ditto, this was testing that you got 204 and an empty array back if you requested a playbook with no instances, why remove the length check

Also I dont know if there is a difference in behavior between listing instances under an *existing* etcd directory (`testPlaybookFull`) and an *empty* etc directory (`testPlaybookEmpty`)

services/instance_test.go
```diff
@@ -41,3 +41,18 @@ func TestShowMissingInstance(t *testing.T) {
 	assert.NotNil(t, err)
 	assert.Empty(t, instance.PlaybookID, "PlaybookID should be empty")
 }
+
+func TestAllWithPlaybookID(t *testing.T) {
```
therealplato: we should add a test case for multiple results

server/server_test.go
cored: @therealplato Right, I'll check it out.

macat:
This is great! +1

server/server_test.go
cored: @therealplato I feel like returning `204` is an over specification just because the json will be empty, why do we also return an http code. In each case from the client point of view it will have to do a check the only thing that maybe you are saving for them is the parsing of the JSON and for the mvp do we really care?

services/instance_test.go
cored: @therealplato This is actually just enforcing that the method exist; those edge cases are tested at the repo level which they belong. Since I'm returning the same error messages from within that layer there's no need to test that in here. And that's the beauty of separating the application in different layers.

PR #25, opened 2016-03-29
Server test cleanup

macat:


broadway/repository_test.go
```diff
@@ -65,3 +65,24 @@ func TestFindByID(t *testing.T) {
...
+
+func TestFindByPlaybookIDNoExistent(t *testing.T) {
+	repo := NewInstanceRepo(&DummyStore{})
+
+	instances, err := repo.FindByPlaybookID("notcreated")
+
+	assert.NotNil(t, err)
+	assert.Empty(t, instances)
+
+	assert.Equal(t, "Saved data for this instance is malformed", err.Error())
```
therealplato: the error here should indicate that there is no playbook with that id

cored: @therealplato That sounds more like a `NotFound` error instead of a problem on the save which is what this error is exposing.

therealplato: what save? isnt this just a lookup of all leaf etcd nodes under `.../notcreated/`? i would expect a NotFound error if I look up a playbook that doesn't exist, or an empty array of instances if I look up an existing playbook with no saved instances.

broadway/repository.go
```diff
@@ -73,3 +74,20 @@ func (ir *InstanceRepo) FindByID(playbookID, ID string) (*Instance, error) {
 	path := "/broadway/instances/" + playbookID + "/" + ID
 	return ir.FindByPath(path)
 }
+
+// FindByPlaybookID finds instances by playbook id
+func (ir *InstanceRepo) FindByPlaybookID(playbookID string) ([]*Instance, error) {
+
+	data := ir.store.Values(fmt.Sprintf("/broadway/instances/%s", playbookID))
```
bobbytables: I'm tempted to think we should move this format string into a const.

broadway/repository.go
```diff
@@ -73,3 +74,20 @@ func (ir *InstanceRepo) FindByID(playbookID, ID string) (*Instance, error) {
...
+	instances := []*Instance{}
```
bobbytables: Why does `Instance` need to be a pointer here?

PR #26, opened 2016-03-30
Deploy with slack command

therealplato:
This PR parses "/broadway deploy foo bar" when sent through Slack and triggers the deployment.


server/server.go
```diff
@@ -200,27 +202,61 @@ func (s *Server) postCommand(c *gin.Context) {
...
+	c.String(code, output)
 	return
 }
 
-func helperRunCommand(text string) (string, error) {
-	return "unimplemented :sadpanda:", nil
+// helperRunCommand takes the plaintext command, minus the leading /broadway
+// trigger, and returns statusCode, message, error for output to the user
+func helperRunCommand(s *Server, text string) (int, string, error) {
+	hint := `/broadway help: This message
```
bobbytables: This should be a constant.

server/server_test.go
```diff
@@ -257,6 +257,23 @@ func TestPostCommandHelp(t *testing.T) {
...
+	if err := os.Setenv(slackTokenENV, testToken); err != nil {
+		t.Fatal(err)
+	}
+	w, server := helperSetupServer()
+	req, _ := http.NewRequest("POST", "/command", nil)
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	form := url.Values{}
+	form.Set("token", testToken)
+	form.Set("command", "/broadway")
+	form.Set("text", "deploy foo")
```
bobbytables: Can't you do this when you initialize the `url.Values` struct?

therealplato: @macat and i switched to this syntax while troubleshooting form problems, I think we were having trouble setting the body properly from the url.Values struct

server/server.go
```diff
@@ -200,27 +204,59 @@ func (s *Server) postCommand(c *gin.Context) {
...
-	c.String(http.StatusOK, output)
+	c.String(code, output)
 	return
 }
 
-func helperRunCommand(text string) (string, error) {
-	return "unimplemented :sadpanda:", nil
+// helperRunCommand takes the plaintext command, minus the leading /broadway
+// trigger, and returns statusCode, message, error for output to the user
+func helperRunCommand(s *Server, text string) (int, string, error) {
```
macat: simplify the name to maybe `runCommand`

macat:
+1

server/server_test.go
bobbytables: Interesting.

server/server.go
```diff
@@ -200,28 +205,59 @@ func (s *Server) postCommand(c *gin.Context) {
...
+
+		_, err := doDeploy(s, commands[1], commands[2])
+		if err != nil {
+			return http.StatusInternalServerError, "Deployment failed", err
+		}
+		msg := fmt.Sprintf("Instance %s/%s deployed", commands[1], commands[2])
+		return http.StatusOK, msg, nil
+	default:
+		return http.StatusNotImplemented, "unimplemented :sadpanda:", nil
+	}
```
bobbytables: I think this ok for _for now_ but we really should be aiming for Open-Closed principle. Right now to add another slash command we have to modify this function, not ideal.

https://en.wikipedia.org/wiki/Open/closed_principle

bobbytables:
:+1:

server/server.go
therealplato: It's probably a good time to refactor it when we are adding CLI, otherwise we will be repeating ourselves

bobbytables:
:+1:

PR #27, opened 2016-03-31
adding instance id, playbook id to template variables

macat:


services/deployment.go
```diff
@@ -28,6 +28,20 @@ func NewDeploymentService(s store.Store, ps map[string]*playbook.Playbook, ms ma
...
 
+func vars(i *Instance) map[string]string {
+	var vs map[string]string
+	for k, v := range i.Vars {
+		vs[k] = v
+	}
+
+	vs["playbook_id"] = i.PlaybookID
+	vs["instance_id"] = i.ID
+	vs["id"] = i.ID
```
poopoothegorilla: is there a reason for having an `instance_id` and an `id`? What are the cases where you would have both different?

macat: Just convenience for the users.

PR #28, opened 2016-03-31
CORE-229: Slack success notification generation

cored:


services/slack_deployment_notifier.go
```diff
@@ -0,0 +1,46 @@
...
+	UserID      string
+	UserName    string
+	Command     string
+	Text        string
+	ResponseUrl string
+}
+
+func (sp *SlackPayload) InstanceInfoFromCommand() (playbookID, id string) {
+	parsedCommand := strings.Split(sp.Command, " ")
+	return parsedCommand[2], parsedCommand[3]
```
therealplato: this will panic with instance out of range if this function gets called with e.g. `/broadway deploy`. see other comment below, you'll check `Text` not `Command`

services/slack_notifier_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	slackPayload := SlackPayload{
+		Token:       "mytoken",
+		TeamID:      "T0001",
+		TeamDomain:  "example",
+		ChannelID:   "C2147483705",
+		ChannelName: "test",
+		UserID:      "U2147483697",
+		UserName:    "Steve",
+		Command:     "/broadway deploy mine pr001",
```
therealplato: this should be just `"/broadway"`, set the `text` field below to `"deploy mine pr001"`

cored: @therealplato Cool, got confuse

services/slack_deployment_notifier.go
cored: @therealplato right another test will confirm that. Even if that's true there's no test yet to add that fix. We are going to add it when we start handling errors.

services/slack_notification.go
```diff
@@ -0,0 +1,30 @@
...
+	return fmt.Sprintf(
+		"%s: %s-%s %s",
+		sn.commandOwner,
+		sn.instance.PlaybookID,
+		sn.instance.ID,
+		sn.instance.Status)
+}
+
+// SlackCommandOwner represents the owner for the slack command
+type SlackCommandOwner string
```
macat: Why do you need a type for this?

services/slack_notification.go
```diff
@@ -0,0 +1,30 @@
...
+
+// NewSlackNotification builds a new slack notification
+func NewSlackNotification(instance *instance.Instance, owner SlackCommandOwner) SlackNotification {
```
macat: How are we going to know who is the owner?

services/slack_notification.go
```diff
@@ -0,0 +1,30 @@
+package services
+
+import (
+	"fmt"
+
+	"github.com/namely/broadway/instance"
+)
+
+// SlackNotification represent a notification
+type SlackNotification struct {
```
macat: Are you planning to add more methods to this struct?

services/slack_notification.go
cored: @macat this is just a fancy wrapper for an string; the idea is to have proper names for the domain models type. 

https://sourcemaking.com/refactoring/smells/primitive-obsession

services/slack_notification.go
cored: @macat from the outer context; the code that is going to use this service will be responsible for instantiating a new owner.

services/slack_notification.go
cored: @macat more methods apart from `String()` ?

macat: Yes

cored: @macat Well, I don't have anything in mind right now but since I'm saving some state for te owner and instance that's wy this is an struct. But I think I will have to change my mind on this one because I need to do a lookup to the instance.

bobbytables:
Simple enough, :+1:

macat:
+1

PR #33, opened 2016-04-04
Load kubernetes configuration

macat:


deployment/config.go
```diff
@@ -0,0 +1,76 @@
...
+
+// KubernetesConfig returns Kubernetes configuration for native Kubernetes
+// environment
+func KubernetesConfig() (*restclient.Config, error) {
+	token, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	if err != nil {
+		return nil, err
+	}
+	return &restclient.Config{
+		Host:        "http://" + os.Getenv("KUBERNETES_SERVICE_HOST") + ":" + os.Getenv("KUBERNETES_PORT_443_TCP_PORT"),
```
bobbytables: We _really_ should not be using environment variables from inside packages, we should pass in the values to this function. All values should be passed from environment variables from the CLI `main` package, not in the package itself.

macat: Thanks, we are going to do this.

poopoothegorilla:
looks good so far :+1:

bobbytables:
:+1:

PR #35, opened 2016-04-05
Consolidate environment variables in one file

therealplato:


deployment/config.go
```diff
@@ -4,18 +4,12 @@ import (
...
 
-var (
-	// EnvServiceHost is the Kubernetes host
-	EnvServiceHost = os.Getenv("KUBERNETES_SERVICE_HOST")
-	// EnvServicePort is the Kubernetes port
-	EnvServicePort = os.Getenv("KUBERNETES_PORT_443_TCP_PORT")
+	"k8s.io/kubernetes/pkg/client/restclient"
 )
 
 func init() {
```
poopoothegorilla: porque?

env/env.go
```diff
@@ -0,0 +1,46 @@
...
+func LoadEnvs() {
+	SlackToken = loadAndWarn("SLACK_VERIFICATION_TOKEN")
+	ServerHost = loadAndWarn("HOST")
+	K8sServiceHost = loadAndWarn("KUBERNETES_SERVICE_HOST")
+	K8sServicePort = loadAndWarn("KUBERNETES_PORT_443_TCP_PORT")
+	K8sNamespace = loadAndWarn("KUBERNETES_NAMESPACE")
+	EtcdHost = loadAndWarn("ETCD_HOST")
+}
+
+func loadAndWarn(key string) (val string) {
```
poopoothegorilla: Do you think we need to explicitly state `AndWarn`? it doesn't look like we will need a regular `load` function... what do you think?

poopoothegorilla: not really a big deal at all though

poopoothegorilla:
+1

env/env.go
therealplato: I named it what it does!
We may want a loadOrFail function for any mandatory variables

PR #36, opened 2016-04-05
Switch to glog

macat:


deployment/manifest_step.go
```diff
@@ -2,8 +2,8 @@ package deployment
 
 import (
 	"errors"
-	"log"
 
+	"github.com/golang/glog"
```
cored: could you add the alias 'log' to this package?

macat: keeping it `glog` makes the code more explicit.

poopoothegorilla:
just a general question. What are our guidelines for when to use `Info`, `Error`, `Warning`, and `Fatal`?

store/etcd.go
```diff
@@ -54,8 +54,7 @@ func (*etcdStore) Values(path string) (values map[string]string) {
 	values = map[string]string{}
 	resp, err := api.Get(context.Background(), path, &etcdclient.GetOptions{Recursive: true})
 	if err != nil {
-		log.Println("Ignoring error getting values:" + path)
-		log.Println(err)
+		glog.Errorf("Ignoring error getting values: %s. Error: %s", path, err.Error())
```
poopoothegorilla: This is just an example of a line i was semi confused on... I would have chosen `Warn` here since it is ignoring the error

macat: This should be a warning, fixing it/

poopoothegorilla:
+1

deployment/manifest_step.go
cored: You already merge it. But I did not get how glog makes the code explicit.

PR #38, opened 2016-04-06
Slack notifications

macat:
- adding new notification package
- triggering notifications after creation and deployment


services/helper_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/namely/broadway/env"
+)
+
+type notificationTestHelper struct {
+	requestBody string
+	ts          *httptest.Server
+}
+
+func newNotificationtestHelper() *notificationTestHelper {
```
poopoothegorilla: not to be too picky but can we change this `newNotificationtestHelper()` to `newNotificationTestHelper()`

services/deployment_test.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 func TestDeployment(t *testing.T) {
+	nt := newNotificationtestHelper()
```
poopoothegorilla: here too just to follow the other naming patterns

poopoothegorilla:
+1

env/env.go
```diff
@@ -27,20 +27,24 @@ var (
...
+	ServerHost = loadw("HOST")
+	K8sServiceHost = loadw("KUBERNETES_SERVICE_HOST")
+	K8sServicePort = loadw("KUBERNETES_PORT_443_TCP_PORT")
+	K8sNamespace = loadf("KUBERNETES_NAMESPACE")
+	EtcdHost = loadw("ETCD_HOST")
+	EtcdPath = loadw("ETCD_PATH")
 }
 
-func load(key string) string {
+func loadw(key string) string {
```
cored: @macat why the change of name for this function?

notification/slack.go
```diff
@@ -0,0 +1,52 @@
...
+	"bytes"
+	"encoding/json"
+	"net/http"
+
+	"github.com/golang/glog"
+	"github.com/namely/broadway/env"
+)
+
+// Field is a Slack message field
+type Field struct {
```
cored: why just don't use the `SlackMessage` name then?

notification/slack.go
```diff
@@ -0,0 +1,52 @@
...
+	Title string `json:"title"`
+	Value string `json:"value"`
+	Short bool   `json:"short"`
+}
+
+// Attachment is Slack message attachment
+type Attachment struct {
```
cored: Why do we need to think about attachment ?

services/instance.go
```diff
@@ -17,7 +20,17 @@ func NewInstanceService(s store.Store) *InstanceService {
 
 // Create a new instance
 func (is *InstanceService) Create(i *instance.Instance) error {
-	return is.repo.Save(i)
+	err := is.repo.Save(i)
+	if err != nil {
+		return err
+	}
+
+	err = sendCreationNotification(i)
```
cored: I was thinking in more of the lines of: 
```
  instanceService := NewInstanceService(store, notifier)  // Notifier should implement the notification interface with the method Send() 
```
In that way we are actually exposing the relationship between the service and the notifications. And taking into account that you mention 'mail notification' at some point that approach will make more sense because the implementation detail will be interchangeable. 

/cc @bobbytables

notification/slack.go
therealplato: for e.g. a red/black/green bar highlighting the bot response

services/instance.go
therealplato: i'd prefer `is.setNotifier()` otherwise it breaks the existing API, and does it really make sense to force the user to have a Notifier if they want to use InstanceService?

services/helper_test.go
macat: fixed

env/env.go
macat: because there are now `loadf` and `loadw` methods

services/instance.go
macat: I agree with creating an abstraction, but only when we actually have more notifiers.

notification/slack.go
macat: https://api.slack.com/incoming-webhooks

therealplato:
+1

env/env.go
cored: @macat right but why two different methods to load environment variables?

services/instance.go
cored: @macat at taht point we are going to miss the higher level interface for the notifiers and the refactor will be bigger. But ok.

env/env.go
macat: `loadw` shows a warning, `loadf` exits if the env var is missing

PR #39, opened 2016-04-06
Properly handle Slack commands

therealplato:
This PR moves processing logic for Slack commands into `services/slack-_command_handler.go`

Callers will invoke `services.BuildSlackCommand()` to get a SlackCommand object.
When `.Execute()` is called on this object, the appropriate action will be taken.


therealplato:
Closed pending @cored refactor

PR #41, opened 2016-04-06
quick fix for kub env

macat:
broadway didn't pick up native kubernetes env variables.


deployment/config.go
```diff
@@ -19,7 +20,8 @@ func IsKubernetesEnv() bool {
 
 	for _, file := range files {
 		fi, err := os.Stat(file)
-		if err != nil || !fi.IsDir() {
+		if err != nil || fi.IsDir() {
```
cored: @macat could you add a test for this please.

PR #44, opened 2016-04-07
quick and dirty slack setvars

therealplato:
This PR sets vars on an existing instance when the appropriate slack command is received by the server.

it does NOT verify that all vars are listed in the instance's playbook
it DOES allow you to set a var to an empty string
it does NOT allow you to set a var or value containing a space character
it does NOT redeploy if the instance is already deployed


server/server.go
```diff
@@ -236,6 +239,39 @@ func doCommand(s *Server, text string) (int, string, error) {
...
+				return http.StatusOK, "Instance Not Found", nil
+			default:
+				glog.Errorf("Problem with is.Show(%s, %s): %s", commands[1], commands[2], err)
+				return http.StatusInternalServerError, "Server Error", err
+			}
+		}
+		vars := make(map[string]string)
+		kvs := commands[3:]
+		for _, kv := range kvs {
+			kv = strings.TrimSuffix(kv, "=")
```
macat: I'd like to suggest another version here. This way values could contain equal (=) characters and save a couple of lines and operations.

```go
tmp := strings.SplitN(kv, "=", 2)
if len(tmp) < 2 || tmp[0] == "" {
  // commandhint
}
```

server/server.go
```diff
@@ -236,6 +239,37 @@ func doCommand(s *Server, text string) (int, string, error) {
...
+		vars := make(map[string]string)
+		kvs := commands[3:]
+		for _, kv := range kvs {
+			tmp := strings.SplitN(kv, "=", 2)
+			if len(tmp) < 2 || tmp[0] == "" {
+				return http.StatusOK, commandHint, nil
+			}
+			vars[tmp[0]] = tmp[1]
+		}
+		i.Vars = vars
```
cored: @therealplato This implementation is wrong; you are updating all the variables without checking if the instance defines that particular variable in the playbook.

macat:
https://github.com/namely/broadway/pull/46

PR #45, opened 2016-04-07
RFC: Reorganize file structure

therealplato:
I'm unhappy with our current file structure, what does the team think of this change?
Our packages except for `env` are been moved into a new `broadway` package.
I envision this `broadway` package as the sole interface used by broadway clients, whether they are importing the broadway package or invoking behavior from CLI or invoking behavior from HTTP.

`thirdparty.go.bak` and `broadwayctl.go.bak` illustrate some ways the broadway package could act as an entry point.

Todo: Consolidate `broadway/*` packages; move server package to top level and make it depend only on the `broadway` package

screen shot 2016-04-07 at 9 33 20 am


macat:
Seen other projects using `pkg` folder for lib packages (https://github.com/coreos/etcd/tree/master/pkg https://github.com/kubernetes/kubernetes/tree/master/pkg).

I'm not a big fan of broadway/broadway style package names. Too long and unnecessary.

Also our project is not big enough to start hiding these folders yet.

poopoothegorilla:
yeh i donna how i feel about this either "github.com/namely/**broadway/broadway**/store"

macat:
Can we close this?

therealplato:
Closing; will improve structure more gradually e.g. #47 
See also my broadway-as-a-package usage idea, vanishing with this branch:
https://gist.github.com/therealplato/75177a442eb7d7a904ad30bfe4c359b8

PR #46, opened 2016-04-07
CORE-215: Update environment variables for an instance

cored:


services/instance.go
```diff
@@ -19,18 +19,21 @@ func NewInstanceService(s store.Store) *InstanceService {
 }
 
 // Create a new instance
-func (is *InstanceService) Create(i *instance.Instance) error {
+func (is *InstanceService) Create(i *instance.Instance) (*instance.Instance, error) {
 	err := is.repo.Save(i)
```
bobbytables: We should probably log something here.

services/instance_test.go
```diff
@@ -59,3 +57,35 @@ func TestAllWithPlaybookID(t *testing.T) {
...
+		Instance           *instance.Instance
+		ExpectedPlaybookID string
+		ExpectedID         string
+		ExpectedVars       map[string]string
+		E                  error
+	}{
+		{
+			"When the instance have all the needed values",
+			&instance.Instance{PlaybookID: "foo", ID: "bar"},
+			"bar",
```
therealplato: i think ExpectedPlaybookID and ExpectedID are switched

services/instance.go
```diff
@@ -19,18 +19,21 @@ func NewInstanceService(s store.Store) *InstanceService {
...
-		return err
+		return nil, err
 	}
+	return createdInstance, nil
+}
 
-	return nil
+// Update an instance
+func (is *InstanceService) Update(instance *instance.Instance) (*instance.Instance, error) {
+	return is.Create(instance)
```
bobbytables: `InstanceService` should probably have an `Update` on it, even if it just calls `Create`. Because in our documentation we should say that it just delegates the call. Reading this threw me off.

services/slack_command_handler.go
```diff
@@ -0,0 +1,64 @@
...
+func (c *setvarCommand) Execute() error {
+	c.Vars = make(map[string]string)
+	kvs := c.args[3:len(c.args)] // from e.g. "setvar foo bar var1=val1 var2=val2"
+	for _, kv := range kvs {
+		kv = strings.TrimPrefix(kv, "=")
+		kv = strings.TrimSuffix(kv, "=")
+		tmp := strings.Split(kv, "=")
+		if len(tmp) != 2 {
+			glog.Warning("Setvar tried to parse badly formatted variable: " + kv)
+			return errors.New("This is not the proper syntax. ex: var1=val1")
```
bobbytables: Errors like this should probably be a `var` for the package.

```go
var ErrInvalidCommandSyntax = errors.New("This is not the proper syntax. ex: var1=val1")
```

services/slack_command_handler_test.go
```diff
@@ -0,0 +1,89 @@
...
+			"When a valid set of arguments been passed",
+			[]string{"setvar", "foo", "bar", "var1=val1"},
+			map[string]string{"var1": "val1"},
+			nil,
+		},
+		{
+			"When an argument text just has a key",
+			[]string{"setvar", "foo", "bar", "var1="},
+			map[string]string{},
+			errors.New("This is not the proper syntax. ex: var1=val1"),
```
bobbytables: So here you can use that exported var in the test a lot easier.

services/instance_test.go
```diff
@@ -59,3 +57,35 @@ func TestAllWithPlaybookID(t *testing.T) {
...
+			"foo",
+			map[string]string{},
+			nil,
+		},
+	}
+
+	for _, testcase := range testcases {
+		createdInstance, _ := instanceService.Create(testcase.Instance)
```
bobbytables: Don't ditch the error please.

services/instance.go
cored: But that's what this code is doing right now. Exposing an update method in
the service and calling it's own create

services/slack_command_handler.go
cored: Yes I thought about thst, too.

services/instance_test.go
```diff
@@ -50,7 +48,7 @@ func TestAllWithPlaybookID(t *testing.T) {
 	service := NewInstanceService(store)
 
 	i := &instance.Instance{PlaybookID: "test", ID: "222"}
-	err := service.Create(i)
+	_, err := service.Create(i)
 	if err != nil {
 		t.Log(err)
 	}
```
bobbytables: Can we update this as a boyscout to include the message on the `assert.Nil` call?

bobbytables: Basically getting rid of the `t.Log` and keeping it on the assert package.

bobbytables:
So I'm a little confused, this doesn't seem to care about slack really. There's no server component or integration level testing for this. The story has "User can set a variable from slack", but this PR hasn't quite accomplished that.

https://namely.atlassian.net/browse/CORE-215

services/instance.go
macat: This should be something like.. maybe... `.Save`?

bobbytables: Oh right, sorry.

services/instance.go
cored: @bobbytables will do that.

services/instance_test.go
cored: Will double check on that @therealplato

services/instance_test.go
cored: @bobbytables I normally do this on tests just because I'm not testing the create. But I guess I can log the error if any.

services/instance_test.go
cored: @bobbytables I don't like that idea, because that will imply making assertions outside of the concern of that particular test case.

cored:
@bobbytables Well, this is a WIP as I mention in the channel I'm working on the web part of it now. But the advantage is that it will be mostly glue code at this point.

instance/repository_test.go
```diff
@@ -69,8 +72,11 @@ func TestFindByID(t *testing.T) {
 func TestFindByPlaybookIDOne(t *testing.T) {
 	repo := NewRepo(store.New())
 	i := &Instance{PlaybookID: "one", ID: "222"}
-	repo.Save(i)
+	err := repo.Save(i)
 
+	if err != nil {
```
bobbytables: Typically you'd do an inline error her, something like this:

```go
if err := repo.Save(i); err != nil {
  t.Error(err)
}
```

But I'd also ask if this could use `assert` package to do this?

services/instance.go
bobbytables: Did you add a log statement?

services/instance.go
```diff
@@ -19,18 +19,25 @@ func NewInstanceService(s store.Store) *InstanceService {
...
+		return i, err
 	}
+	return i, nil
+}
 
-	return nil
+// Update an instance
+func (is *InstanceService) Update(i *instance.Instance) (*instance.Instance, error) {
+	err := is.repo.Save(i)
+	if err != nil {
```
bobbytables: Same sorta deal here with an inline `if err := ...; err != nil {`

services/instance_test.go
bobbytables: How is this different than including the log? It's just more concise really:

```go
assert.Nil(t, err, "err should be nothing")
```

thoughts?

services/slack_command_handler.go
```diff
@@ -0,0 +1,88 @@
...
+	return "That's not the proper syntax. ex: var1=val1"
+}
+
+type setvarCommand struct {
+	args []string
+	is   *InstanceService
+}
+
+func (c *setvarCommand) Execute() (string, error) {
+	kvs := c.args[3:len(c.args)] // from e.g. "setvar foo bar var1=val1 var2=val2"
```
therealplato: i believe this will blow up if you invoke with too few args e.g. `/broadway setvar`

services/slack_command_handler.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (c *helpCommand) Execute() (string, error) {
+	return `/broadway help: This message
+/broadway deploy myPlaybookID myInstanceID: Deploy a new instance`, nil
+}
+
+// BuildSlackCommand takes a string and some context and creates a SlackCommand
+func BuildSlackCommand(payload string, is *InstanceService) SlackCommand {
+	terms := strings.Split(payload, " ")
+	switch terms[0] {
```
therealplato: i think this will blow up if payload is empty string

services/slack_command_handler.go
```diff
@@ -0,0 +1,88 @@
...
+	i, err := c.is.Show(c.args[1], c.args[2])
+	var commandMsg string
+	if err != nil {
+		glog.Warningf("Cannot setvars for not found instance %s/%s\n", c.args[1], c.args[2])
+		return "", err
+	}
+	for _, kv := range kvs {
+		kv = strings.TrimPrefix(kv, "=")
+		kv = strings.TrimSuffix(kv, "=")
+		tmp := strings.Split(kv, "=")
```
bobbytables: Could we use `url.Values` here to do the parsing? It seems like it might be a good solution, (https://golang.org/pkg/net/url/#ParseQuery)

```go
values, err := url.ParseQuery(kv)
```

`values` would just a map now, you'd just have to convert it to whatever `i.Vars` type is.

services/slack_command_handler_test.go
```diff
@@ -0,0 +1,105 @@
...
+		Scenario string
+		Args     string
+		Expected string
+		E        error
+	}{
+		{
+			"When passing help command",
+			"help",
+			`/broadway help: This message
+/broadway deploy myPlaybookID myInstanceID: Deploy a new instance`,
```
bobbytables: We should probably move this string to a const, this is going to change as we add more commands.

services/slack_command_handler.go
macat: Isn't that for `var=value&var2=value` type of query strings?

services/instance.go
macat: I like having them in separate lines, more readable.

services/slack_command_handler.go
```diff
@@ -0,0 +1,88 @@
+package services
```
macat: rename this file to `command.go`

macat: or maybe `slack.go`

services/slack_command_handler.go
macat: fix this.

services/slack_command_handler.go
macat: You can use SplitN here

therealplato: SplitN implementation https://github.com/namely/broadway/pull/44/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR259

services/slack_command_handler.go
```diff
@@ -0,0 +1,88 @@
...
+		return "", err
+	}
+	return commandMsg, nil
+}
+
+// Help slack command
+type helpCommand struct{}
+
+func (c *helpCommand) Execute() (string, error) {
+	return `/broadway help: This message
```
therealplato: can you use my updated string and const from here? https://github.com/namely/broadway/pull/44/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR31

services/instance_test.go
cored: @bobbytables The thing is that I normally check the fault case related to the test case. The service call is just to be consistent across the layers and keep using the same entry points to the domain model. I really don't care much about asserting that this throws an error because this is not the scope of this particular test case. That's why we have an instance_test.go

services/slack_command_handler.go
bobbytables: What's to say you can't just use it multiple times...?

services/slack_command_handler.go
bobbytables: Actually this one will be fine. A split string always has at least one entry in the slice.

services/instance_test.go
bobbytables: I still think what I suggested and what you typed align perfectly but lets move on.

services/instance_test.go
bobbytables: The reason I like testing err in tests is because if there is one, `createdInstance` (or whatever is returned) is typically `nil`. Which means on line 83 you'd get a nil pointer exception, which is harder to initially digest.

cored: @bobbytables I see; but wouldn't be better to add a `.Fail` call so the testcase stops at that moment?

bobbytables: Yeah I usually would do an immediate fail, there's no point to carry on at that point.

services/slack_command_handler.go
bobbytables: I don't think there's a gain there @macat

services/instance_test.go
cored: Good will switch it.

services/slack_command_handler.go
macat: The package can still be `services`

services/slack_command_handler.go
cored: @bobbytables I think you are right we should consider this parsing mechanism it will return the proper type for the an `Instance.Vars`

bobbytables:
:shipit:

services/slack.go
```diff
@@ -0,0 +1,92 @@
...
+		glog.Warningf("Cannot setvars for not found instance %s/%s\n", c.args[1], c.args[2])
+		return "", err
+	}
+	for _, kv := range kvs {
+		tmp := strings.SplitN(kv, "=", 2)
+		if len(tmp) != 2 {
+			glog.Warning("Setvar tried to parse badly formatted variable: " + kv)
+			return "", &InvalidSetVar{}
+		}
+		if _, ok := i.Vars[tmp[0]]; ok {
```
macat: Instance might not have those variable set. We should check the `playbook.Vars`.

server/server_test.go
```diff
@@ -282,6 +282,28 @@ func TestPostCommandHelp(t *testing.T) {
 	assert.Equal(t, http.StatusOK, w.Code, "Expected /broadway help to be 200")
 	assert.Contains(t, w.Body.String(), "/broadway", "Expected help message to contain /broadway")
 }
+
```
therealplato: you can probably drop in my additional tests from https://github.com/therealplato/broadway/blob/6296523ba8174a7ce972da56133795161ceb68ac/server/server_test.go#L242-L313

services/slack.go
```diff
@@ -0,0 +1,104 @@
...
+			if playbookVar == name {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// CommandHints slack commands help hints
+const CommandHints = `/broadway help: This message
```
therealplato: you missed the setvar hint
```
const commandHint = `Usage:
/broadway help: This message
/broadway deploy myPlaybookID myInstanceID: Deploy a new instance
/broadway setvar myPlaybookID myInstanceID var1=val1 githash=8ad33dad env=prod ...
`
```

server/server_test.go
therealplato: actually nvm, i see your tests covering "works for 1", "works for blank value" cases below. you could still test "missing instance outputs not found" and pass 2 k/v pairs instead of 1 in your "defines a variable"

cored: But I'm not passing instance information for help and in that case the
factory will select a command if none is found the help will be returned so
no need to test thst.

services/instance_test.go
bobbytables: > Good will switch it.

Did you switch this?

therealplato: @bobbytables @cored i switched this in #49

PR #47, opened 2016-04-08
Consolidate playbook+manifest into deployment package

therealplato:
This PR combines playbook and manifest packages into deployment package.

Tests are green but we could do more:

* Combine `services/manifest.go` into `services/deployment.go`
* Create a `context` object to keep track of available playbooks and manifests throughout the app (would help for checking setvar input against playbooks)
* Add `Task.ManifestSteps []ManifestStep` and `Task.PodManifestStep PodManifestStep` to our Task struct


bobbytables:
Other advantages of this is now we can un-export certain things that only this `deployment` package cares about.

bobbytables:
:+1:

PR #48, opened 2016-04-08
Implement Bearer authentication token

therealplato:
This PR protects all routes except `/command` with an authentication token.

Gin will look at the `Authentication` header and expect to see something like `Bearer 12345`.
It will compare the token with `env.AuthBearerToken` which defaults to the value of the env var `BROADWAY_AUTH_TOKEN`.


server/server.go
```diff
@@ -80,12 +80,16 @@ func (s *Server) Init() {
 func (s *Server) setupHandlers() {
 	s.engine = gin.Default()
 	gin.SetMode(gin.ReleaseMode) // Comment this to use debug mode for more verbose output
+	// Define routes:
+	s.engine.POST("/command", s.postCommand)
+	s.engine.GET("/command", s.getCommand)
+	// Protect subsequent routes with middleware:
+	s.engine.Use(authMiddleware)
```
macat: This is awesome. I didn't know you can use a middleware only for some handlers.

macat:
+1

poopoothegorilla:
+1

macat:
MR STORY EATER

PR #49, opened 2016-04-11
Fix/instance varcreation

therealplato:
This PR adds `vars` validation for instance creation and setting vars for an existing instance. Both will fail if you pass in a var that is not present in the instance's playbook.

Other changes: 

* Instance creation now fails if the referenced `PlaybookID` is missing
* `BROADWAY_MANIFESTS_PATH` and `BROADWAY_PLAYBOOKS_PATH` env vars have been added
* Hardcoded manifest and playbook folder locations have been replaced with these env vars
* The `test` container sets these env vars set to `examples/manifests` and `examples/playbooks` respectively
* All tests that need to manipulate instances now use `PlaybookID: "helloplaybook"`
* `helloplaybook` has been added to the examples folder, it declares two vars: `bird` and `word`


deployment/manifest.go
```diff
@@ -31,3 +38,33 @@ func (m *Manifest) Execute(vars map[string]string) string {
 	}
 	return b.String()
 }
+
+// SetManifestRoot ensures a folder exists, then sets ManifestRoot to that
+// folder.
+func SetManifestRoot(newRoot string) error {
```
cored: @therealplato Why is this related?

deployment/manifest.go
```diff
@@ -31,3 +38,33 @@ func (m *Manifest) Execute(vars map[string]string) string {
...
+	if _, err := os.Stat(newRoot); err != nil {
+		return err
+	}
+	env.ManifestsPath = newRoot
+	return nil
+}
+
+// ManifestsPresent iterates through the Manifests and PodManifest items on a
+// task, and checks that each represents a file on disk
+func (t Task) ManifestsPresent() error {
```
cored: Also this?

env/env.go
```diff
@@ -32,17 +33,41 @@ var (
...
+
+var (
+	cwd                  string
+	defaultManifestsPath string
+	defaultPlaybooksPath string
 )
 
 // LoadEnvs sets env.* variables to their OS-provided value
 func LoadEnvs() {
+	ManifestsPath = loadw("BROADWAY_MANIFESTS_PATH")
```
bobbytables: We probably a `sync.Once` in here so that this can't be called again.

server/server.go
```diff
@@ -70,6 +70,7 @@ func (s *Server) Init() {
 	}
 
 	s.playbooks, err = deployment.LoadPlaybookFolder("playbooks/")
```
cored: @therealplato Why don't you change this to use the new environment variable?

therealplato: good catch

services/instance.go
```diff
@@ -19,17 +20,57 @@ func NewInstanceService(s store.Store) *InstanceService {
 	return &InstanceService{repo: instance.NewRepo(s)}
 }
 
+// PlaybookNotFound indicates a problem due to Broadway not knowing about a
+// playbook
+type PlaybookNotFound struct {
```
cored: I like this.

services/instance.go
```diff
@@ -19,17 +20,57 @@ func NewInstanceService(s store.Store) *InstanceService {
...
+
+func (e *InvalidVar) Error() string {
+	return fmt.Sprintf("Playbook %s does not declare a var named %s\n", e.playbookID, e.key)
+}
+
 // Create a new instance
 func (is *InstanceService) Create(i *instance.Instance) (*instance.Instance, error) {
-	glog.Info("Instance Service: Create")
+	glog.Infof("InstanceService attempting to create %s/%s...\n", i.PlaybookID, i.ID)
+	pb, ok := deployment.AllPlaybooks[i.PlaybookID]
```
cored: You can switch this call to be inline. 
`if pb, ok := deploymemt..; ok { }`

deployment/manifest.go
therealplato: i moved this out of playbook.go while refactoring, i think it can now be removed since e.g. tests can manipulate `env.ManifestsPath`

deployment/manifest.go
therealplato: This is used in playbook validation

services/instance.go
```diff
@@ -19,17 +20,57 @@ func NewInstanceService(s store.Store) *InstanceService {
...
+	if ok != true {
+		return nil, &PlaybookNotFound{i.PlaybookID}
+	}
+	// Set all valid vars to empty string:
+	vars := make(map[string]string)
+	for _, pv := range pb.Vars {
+		vars[pv] = ""
+	}
+	// Look for invalid vars on new instance, and set values:
```
cored: What does an invalid mean in this context? Why do we need to set them?

services/instance.go
```diff
@@ -19,17 +20,57 @@ func NewInstanceService(s store.Store) *InstanceService {
...
+	i.Vars = vars
 	err := is.repo.Save(i)
 	if err != nil {
 		return nil, err
 	}
-	err = sendCreationNotification(i)
-	if err != nil {
-		return i, err
-	}
+	/*
```
cored: Remove this commented code.

services/instance_test.go
```diff
@@ -8,47 +8,78 @@ import (
...
+	service := NewInstanceService(store)
+
+	i := &instance.Instance{PlaybookID: "vanishing-pb", ID: "gone"}
+	_, err := service.Create(i)
+
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "playbook vanishing-pb is missing")
+}
+
+func TestCreateInstanceWithIncorrectVars(t *testing.T) {
```
cored: Can you add this or remove it?

services/instance_test.go
```diff
@@ -8,47 +8,78 @@ import (
...
+	//todo: set the wrong vars
+}
+
+/*
+func TestCreateInstanceNotification(t *testing.T) {
```
cored: Remove this commented code. Please.

services/instance.go
therealplato: @macat actually if you have time can you uncomment this, and the associated test, and see if you can reproduce my `net` error in sending notification?

services/instance_test.go
therealplato: yep will add these tests

services/instance_test.go
```diff
@@ -8,47 +8,78 @@ import (
...
-	assert.Equal(t, "test", createdInstance.PlaybookID)
-	assert.Equal(t, instance.StatusNew, createdInstance.Status)
+
 	assert.Contains(t, nt.requestBody, "created")
+	assert.Contains(t, nt.requestBody, "helloplaybook")
+	assert.Contains(t, nt.requestBody, "TestCreateInstanceNotification")
+}
+*/
+
+func TestCreateInstance(t *testing.T) {
```
cored: I think we can use the test table idea for this method.

services/instance_test.go
```diff
@@ -8,47 +8,78 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestCreateInstance(t *testing.T) {
+func TestCreateInstanceFromMissingPlaybook(t *testing.T) {
```
cored: @therealplato Is there a test for the create that stress the code for 'invalid variables' ? I can't see it in here.

services/instance.go
therealplato: if i did that, wouldn't pb be out of scope after the if block? i need to use it after the if block

services/slack_test.go
```diff
@@ -26,56 +21,56 @@ func TestSetvarExecute(t *testing.T) {
 		E            error
 	}{
 		{
-			"When an instance and playbook defines a variable",
-			"setvar foo bar var1=val1",
-			&instance.Instance{PlaybookID: "foo", ID: "bar", Vars: map[string]string{"var1": "val2"}},
-			map[string]*deployment.Playbook{"foo": &deployment.Playbook{ID: "foo", Vars: []string{"var1"}}},
-			map[string]string{"var1": "val1"},
-			"Instance foo bar updated it's variables",
+			"When an instance sets a new variable value from the playbook",
```
cored: This is test is what suggest me that we need to inject the playbooks into the `.Create` method in the instance service as a parameter. I know we discuss this but the refactoring could be part of the same PR. What do you think?

services/instance.go
cored: @therealplato Try it :-)

poopoothegorilla:
+1 looks good :)

macat:
+1

PR #50, opened 2016-04-12
CORE-240: Instance deletion

cored:


services/instance_test.go
```diff
@@ -140,3 +140,27 @@ func TestUpdate(t *testing.T) {
...
+	is := NewInstanceService(store.New())
+
+	i := &instance.Instance{PlaybookID: "foo", ID: "bar"}
+
+	createdInstance, err := is.Create(i)
+	if err != nil {
+		t.Fail()
+	}
+	err = is.Delete(createdInstance)
+	assert.Nil(t, err, "When existent instance")
```
bobbytables: Probably would be better worded as "when interface exists"

bobbytables:
I like this, the abstractions are clearer the more we use them.

services/instance_test.go
cored: changing it.

server/server.go
```diff
@@ -87,6 +87,7 @@ func (s *Server) setupHandlers() {
 	s.engine.GET("/instances/:playbookID", s.getInstances)
 	s.engine.GET("/status/:playbookID/:instanceID", s.getStatus)
 	s.engine.POST("/deploy/:playbookID/:instanceID", s.deployInstance)
+	s.engine.DELETE("/:playbookID/:instanceID", s.deleteInstance)
```
therealplato: i suggest `DELETE instances/:playbookID/:instanceID`

cored: @therealplato yes, you are right.

macat:
+1

instance/repository.go
```diff
@@ -92,3 +93,8 @@ func (ir *Repo) FindByPlaybookID(playbookID string) ([]*Instance, error) {
 
 	return instances, nil
 }
+
+// Delete remove an instance
```
poopoothegorilla: `remove` vs `removes`

instance/repository_test.go
```diff
@@ -103,3 +103,14 @@ func TestFindByPlaybookIDNoExistent(t *testing.T) {
 
 	assert.Equal(t, "Saved data for this instance is malformed", err.Error())
 }
+
+func TestDelete(t *testing.T) {
+	repo := NewRepo(store.New())
+	i := &Instance{PlaybookID: "anewone", ID: "withid"}
+	err := repo.Save(i)
+	if err != nil {
+		t.Fail()
```
poopoothegorilla: do we care about displaying the error if it fails?

instance/repository_test.go
```diff
@@ -103,3 +103,14 @@ func TestFindByPlaybookIDNoExistent(t *testing.T) {
 
 	assert.Equal(t, "Saved data for this instance is malformed", err.Error())
 }
+
+func TestDelete(t *testing.T) {
```
poopoothegorilla: are there any other checks we might want to perform at this level? it looks like we are just testing for errors

server/server.go
```diff
@@ -264,3 +265,17 @@ func (s *Server) deployInstance(c *gin.Context) {
 	}
 	c.JSON(http.StatusOK, i)
 }
+
+func (s *Server) deleteInstance(c *gin.Context) {
+	is := services.NewInstanceService(s.store)
+	ii, err := is.Show(c.Param("playbookID"), c.Param("instanceID"))
+	if err != nil {
+		c.JSON(http.StatusNotFound, NotFoundError)
```
poopoothegorilla: Is a `NotFoundError` the only error type we expect back? What about `MalformedSavedData` when the un-marshal occurs?

therealplato: what if its some other error?

poopoothegorilla: Or do we only want to send back certain errors?

therealplato: jinx

server/server.go
```diff
@@ -264,3 +265,17 @@ func (s *Server) deployInstance(c *gin.Context) {
...
+		return
+	}
+	err = is.Delete(ii)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, InternalError)
```
therealplato: you need to add a return here otherwise it'll smush together lines 278 and 280 in the response

server/server_test.go
```diff
@@ -330,3 +352,35 @@ func TestDeployMissing(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Contains(t, errorResponse["error"], "Not Found")
 }
+
+func TestDeleteWhenExistentInstance(t *testing.T) {
```
therealplato: use Extant or Existing for better grammar

server/server_test.go
```diff
@@ -330,3 +352,35 @@ func TestDeployMissing(t *testing.T) {
...
+		Status:     instance.StatusDeployed}
+	is := services.NewInstanceService(store.New())
+	_, err := is.Create(testInstance1)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	req, w := testutils.DeleteRequest(
+		t,
+		fmt.Sprintf("/instances/%s/%s", testInstance1.PlaybookID, testInstance1.ID),
```
therealplato: ^ nice

server/server_test.go
```diff
@@ -295,6 +296,27 @@ func TestSlackCommandSetvar(t *testing.T) {
 	assert.Equal(t, http.StatusOK, w.Code, "Expected slack command to be 200")
 }
 
+func TestSlackCommandDelete(t *testing.T) {
+	env.SlackToken = testToken
+	w, server := helperSetupServer()
+	req, _ := http.NewRequest("POST", "/command", nil)
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
poopoothegorilla: just curious is there any benefit of using `Add` versus `Set`?

server/server_test.go
```diff
@@ -330,3 +352,35 @@ func TestDeployMissing(t *testing.T) {
...
+func TestDeleteWhenExistentInstance(t *testing.T) {
+	testInstance1 := &instance.Instance{
+		PlaybookID: "helloplaybook",
+		ID:         "TestGetStatusWithGoodPath",
+		Status:     instance.StatusDeployed}
+	is := services.NewInstanceService(store.New())
+	_, err := is.Create(testInstance1)
+	if err != nil {
+		t.Error(err)
+		return
```
poopoothegorilla: is there a reason you chose to use `return` in this case?

server/server.go
cored: There won't be any unmarshalling when deleting.

server/server_test.go
```diff
@@ -330,3 +352,35 @@ func TestDeployMissing(t *testing.T) {
...
+	}
+	req, w := testutils.DeleteRequest(
+		t,
+		fmt.Sprintf("/instances/%s/%s", testInstance1.PlaybookID, testInstance1.ID),
+	)
+
+	req = auth(req)
+	makeRequest(req, w)
```
poopoothegorilla: is there a difference between choosing `makeRequest` vs  using `HelperSetupServer` and `ServeHTTP`?

services/instance.go
```diff
@@ -98,6 +98,16 @@ func (is *InstanceService) AllWithPlaybookID(playbookID string) ([]*instance.Ins
 	return is.repo.FindByPlaybookID(playbookID)
 }
 
+// Delete remove an instance
```
poopoothegorilla: `remove` to `removes` maybe?

services/instance_test.go
```diff
@@ -129,7 +129,7 @@ func TestUpdate(t *testing.T) {
 	for _, testcase := range testcases {
 		createdInstance, err := instanceService.Create(testcase.Instance)
 		if err != nil {
-			t.Error(err)
+			t.Fail()
```
poopoothegorilla: show error maybe?

services/slack.go
```diff
@@ -93,12 +93,43 @@ func (c *helpCommand) Execute() (string, error) {
 	return commandHints, nil
 }
 
+// InvalidDelete definition of syntax error
+type InvalidDelete struct{}
+
+func (e *InvalidDelete) Error() string {
+	return ""
```
poopoothegorilla: do we want this as an empty string?

services/slack.go
```diff
@@ -93,12 +93,43 @@ func (c *helpCommand) Execute() (string, error) {
...
+		return "", &InvalidDelete{}
+	}
+	iToDelete, err := c.is.Show(c.args[1], c.args[2])
+
+	if err != nil {
+		return "", err
+	}
+	err = c.is.Delete(iToDelete)
+	if err != nil {
+		return "", err
```
poopoothegorilla: I am looking at line 61 for setVarCommand... Why does that return a string as well as an error?

server/server.go
poopoothegorilla: I might be misunderstanding this, but to me it looks like `Show` does unmarshal since it calls `FindByID` and then `FindByPath` https://github.com/namely/broadway/blob/b76b675ad7bd359f0de5cc94d37d2abbe37dcfbc/instance/repository.go#L66

therealplato: good point, why does `is.Delete()` take an instance while `is.Show()` takes ids? should we invoke as `is.Delete(pID, iID)`?

cored: I think you are right on that the ip could be more generic, as Issac
mentions I will give it a little bit of thought.

poopoothegorilla:
+1 still works we can make a pr for changes later

therealplato:
+1

cored:
I'm going to merge it and do individual refactorings.

PR #51, opened 2016-04-12
Rename pod; fix missing manifests dir

therealplato:
This PR fixes a bug where deployments can't find the manifests folder because the server still had a hardcoded reference to `manifests/`


deployment/kubernetes.go
```diff
@@ -89,6 +90,10 @@ func (d *KubernetesDeployment) steps() ([]TaskStep, error) {
 	for _, task := range d.Playbook.Tasks {
 		if task.PodManifest != "" {
 			m := d.Manifests[task.PodManifest]
+			glog.Info("****************")
```
macat: We don't need these

PR #53, opened 2016-04-14
Instance deletion additions

poopoothegorilla:
some minor code changes


therealplato:
why do we see @cored's commits here - because they got squashed in the last PR?

PR #67, opened 2016-05-13
Add tests and implementation for custom create and deploy messages

therealplato:
This PR adds fields to playbook yml:

```
messages:
  created: a go {{.Template}}
  deployed: another {{ .Template }}
```

If these are invalid templates, the playbook will fail to validate.
If the fields are missing, the current default message is shown.
If the fields are present, both the current default and the custom message are shown.

Attachments for create and deploy now have green bars in Slack.


deployment/playbook_test.go
```diff
@@ -22,6 +22,9 @@ vars:
   - version
   - assets_version
   - owner
+messages:
+  created: "help I'm stuck in a test {{ .ID }}"
+  deployed: "help I'm still stuck in a test {{ .ID }}"
```
macat: 😃

services/deployment.go
```diff
@@ -1,8 +1,10 @@
 package services
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"html/template"
```
therealplato: `text/template` ... goimports likes to change this on me

deployment/playbook.go
```diff
@@ -60,6 +62,12 @@ func (p *Playbook) Validate() error {
 	if len(p.Tasks) == 0 {
 		return errors.New("Playbook requires at least 1 task")
 	}
+	for key, value := range p.Messages {
+		_, err := template.New(key).Parse(value)
+		if err != nil {
+			return fmt.Errorf("Playbook had an invalid message template: \"%s\"", value)
```
bobbytables: One thing that I love that the go stdlib's do is that they include the package name in the error:

```go
fmt.Errorf("broadway/playbook: ...")
```

What do you think about maybe doing that here?

bobbytables:
Other than that optional comment I think this looks great. 👍

cored:
+1

services/deployment.go
macat: can you update this?

PR #70, opened 2016-05-17
Use vars method to generate message

macat:


therealplato:
@macat how was this managing to hit the panic? I thought the validation at startup would be good enough to make sure the playbooks had valid templates

macat:
I created a playbook that used `.playbook_id` variable, which resulted in an error and broadway crashed at the panic line.

PR #71, opened 2016-05-18
[WIP] First draft of plugins

macat:
This PR is implementing an external plugin system using GRPC. Admins will be able to set external plugins extending capabilities of Broadway.

With the current implementation an external service will be able to receive event notifications about instances. This way we can move Slack notifications to an external Plugin. I'm also planning to add a plugin for the new LED board.


plugins/events.pb.go
```diff
@@ -0,0 +1,185 @@
...
+
+var InstanceStatus_name = map[int32]string{
+	0: "NEW",
+	1: "DEPLOYING",
+	2: "ERROR",
+	3: "DEPLOYED",
+	4: "DELETING",
+	5: "DELETED",
+}
+var InstanceStatus_value = map[string]int32{
```
cored: So cool!!!

therealplato:
If we moved k8s deployment to a plugin then we could build and test broadway an order of magnitude faster

macat:
@therealplato so true! I'll draft  a protofile for showing  how that could look  like.

therealplato:
#92

PR #78, opened 2016-05-19
Add timestamp to created instances

### LINKED ISSUE #77: Store created time when creating instance
Store unix timestamp at the creation.

therealplato:
Resolves #77


instance/instance.go
```diff
@@ -2,17 +2,18 @@ package instance
...
 // Instance entity
 type Instance struct {
-	PlaybookID string            `json:"playbook_id" binding:"required"`
-	ID         string            `json:"id"`
-	Created    string            `json:"created"`
-	Vars       map[string]string `json:"vars"`
-	Status     `json:"status"`
+	PlaybookID  string            `json:"playbook_id" binding:"required"`
+	ID          string            `json:"id"`
+	CreatedTime string            `json:"created_time"`
```
macat: I think we shouldn't change the name here. `Created` is fine. I'd change the type to `int` though.

services/instance.go
```diff
@@ -94,6 +95,10 @@ func (is *InstanceService) Create(i *instance.Instance) (*instance.Instance, err
 	}
 
 	i.Vars = vars
+
+	// Add a timestamp
+	i.CreatedTime = time.Now().UTC().Format(instance.TimeFmt)
```
macat: We could just store this as an integer, so we don't have to parse date strings.

instance/instance.go
therealplato: I changed it because I thought "Created" was ambiguous, I can imagine values like "yes"/"failed due to x"

services/instance.go
therealplato: I used the formatted string with the expectation that SRE's might be looking at the JSON output and to give them something readable

macat: It's easier to work with an integer even if it's in JSON. You can easily convert it to a readable number

macat: *readable date

instance/instance.go
macat: Yeah, but I want to keep it simple

macat: I'm fine with changing it to `created_time`

therealplato:
@bobbytables suggests adding expiration/ttl alongside this in the same pr

macat:
This is great! +!

PR #81, opened 2016-05-21
New update logic: Comparing resource objects before updating them

macat:
If existing resources are identical to the updated version, Broadway will skip deployment.


therealplato:
+1
Before this PR, were the old resources fully cleaned up after being deleted?

PR #83, opened 2016-05-23
Updating instance not change status

### LINKED ISSUE #72: POST /instances should not overwrite the Instance status
CI is calling POST /instances always, but the current implementation overwrites the instance status to "new". We should keep the instance status if the instance existed before the update.

cored:
Fix #72

### DIFF
--- a/services/instance.go
+++ b/services/instance.go
@@ -99,14 +99,18 @@ func (is *InstanceService) Create(i *instance.Instance) (*instance.Instance, err
 	// This is seconds resolution, multiply by 1000 for javascript ms timestamp:
 	i.Created = time.Now().Unix()
 
+	foundInstance, err := is.repo.FindByID(i.PlaybookID, i.ID)
+	if err == nil {
+		i.Status = foundInstance.Status
+	}
 	err = is.repo.Save(i)
 	if err != nil {
 		return nil, err
 	}
-	// err = sendCreationNotification(i)
-	// if err != nil {
-	// 	return nil, err
-	// }
+	err = sendCreationNotification(i)
+	if err != nil {
+		return nil, err
+	}
 	return i, nil
 }
 
--- a/services/instance.go
+++ b/services/instance.go
@@ -103,10 +103,10 @@ func (is *InstanceService) Create(i *instance.Instance) (*instance.Instance, err
 	if err != nil {
 		return nil, err
 	}
-	err = sendCreationNotification(i)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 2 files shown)


services/instance.go
```diff
@@ -94,6 +94,10 @@ func (is *InstanceService) Create(i *instance.Instance) (*instance.Instance, err
 	}
 
 	i.Vars = vars
+	foundInstance, err := is.repo.FindByID(i.PlaybookID, i.ID)
+	if err == nil {
+		i.Status = foundInstance.Status
```
poopoothegorilla: if you are not using `foundInstance` outside of this block you might want to consider scoping it inside the if...

    if foundInstance, err := ... ; err == nil {
    }

services/instance_test.go
```diff
@@ -49,6 +49,25 @@ func TestCreateInstanceNotification(t *testing.T) {
...
+	store := store.New()
+	is := NewInstanceService(store)
+
+	i := &instance.Instance{PlaybookID: "helloplaybook", ID: "TestCreateInstance"}
+	ii, err := is.Create(i)
+	assert.Nil(t, err)
+	ii.Status = instance.StatusDeployed
+	updatedInstance, err := is.Update(ii)
+	if err != nil {
+		t.Fatal(err)
```
poopoothegorilla: why switch from using this and the testify?

cored: @poopoothegorilla I'm still using testify for this test; but to me the error case is not relevant to this test case so that's why I didn't do an assertion in here.

poopoothegorilla: I understand what you mean, but I feel like if we are using testify for our tests and then we start using `t.Fatal(err)` some places it becomes alittle unstandardized. What if we used `require.Nil(err)`? `require` still ends up calling FailNow like Fatal.

cored: Sounds good to me @poopoothegorilla

services/instance_test.go
```diff
@@ -85,6 +104,10 @@ func TestCreateInstance(t *testing.T) {
 	is := NewInstanceService(store)
 
 	i := &instance.Instance{PlaybookID: "helloplaybook", ID: "TestCreateInstance"}
+	err := is.Delete(i)
+	if err != nil {
+		t.Fail(err)
```
macat: `services/instance_test.go:109: too many arguments in call to t.common.Fail`

macat:
This PR is not fixing our problem which is that the current implementation is overwriting the instance. I'm going to open another PR that'll try to merge variables and keep the instance status if it's an update.

Issue #82, opened 2016-05-23
Adding basic template functions


macat:
We should support basic string manipulations in manifest templates.

Adding the basic functions from `strings` would be enough.

PR #84, opened 2016-05-23
Template functions in manifest templates

macat:
resolves https://github.com/namely/broadway/issues/82

### DIFF
--- a/deployment/manifest_test.go
+++ b/deployment/manifest_test.go
@@ -7,10 +7,30 @@ import (
 )
 
 func TestExecute(t *testing.T) {
-	m, err := NewManifest("test", `{{ .test }}`)
-	assert.Nil(t, err)
+	cases := []struct {
+		name     string
+		template string
+		vars     map[string]string
+		expected string
+	}{
+		{
+			name:     "simple",
+			template: `{{ .test }}`,
+			vars:     map[string]string{"test": "hello!"},
+			expected: "hello!",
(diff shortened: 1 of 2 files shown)


deployment/manifest_test.go
```diff
@@ -7,10 +7,30 @@ import (
...
+		name     string
+		template string
+		vars     map[string]string
+		expected string
+	}{
+		{
+			name:     "simple",
+			template: `{{ .test }}`,
+			vars:     map[string]string{"test": "hello!"},
+			expected: "hello!",
```
poopoothegorilla: It is alittle difficult to tell the different test cases without a `scenario` field or something similar.. is the `name` field supposed to be this as well?

poopoothegorilla:
+1

therealplato:
@macat can you add a test for `datetime`? it returns a Time object and i'm curious how that casts to a string for templating

deployment/manifest_test.go
macat: that's how we named cases in other files

macat: Sorry, I'm wrong. I'll change it.

macat: We have examples for both. I'm fine with scenario

macat:
It implements Stringer interface. So it'll return this. https://golang.org/src/time/format.go?s=12854:12883#L407

PR #87, opened 2016-05-26
Create Or Update instance

### LINKED ISSUE #72: POST /instances should not overwrite the Instance status

macat:
Fixes #72

### DIFF
--- a/services/instance.go
+++ b/services/instance.go
@@ -79,8 +79,6 @@ func (is *InstanceService) CreateOrUpdate(i *instance.Instance) (*instance.Insta
 	}
 
 	existing, err := is.repo.FindByID(i.PlaybookID, i.ID)
-	glog.Info("--------")
-	glog.Info(existing)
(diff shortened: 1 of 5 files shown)


services/instance.go
```diff
@@ -54,56 +57,78 @@ func (e *InvalidID) Error() string {
...
-	sanitizer, err := regexp.Compile(`[^a-zA-Z0-9\-]`)
-	if err != nil {
-		panic(err)
-	}
-	validator, err := regexp.Compile(`^[a-zA-Z0-9\-]{1,253}$`)
-	if err != nil {
-		panic(err)
-	}
-	match := validator.FindStringIndex(i.ID)
+func validateID(id string) error {
```
cored: :-)

cored:
:+1:

PR #88, opened 2016-05-30
gofmt -s

shawnps:


cored:
+1

macat:
We are using go 1.6, no need for explicitly declaring the object types there. Closing this.

shawnps:
@macat maybe I'm misunderstanding but isn't this removing the explicit declaration of `&deployment.Playbook`? I don't think it's related to using Go 1.6

shawnps:
@macat in any case just for reference: https://golang.org/cmd/gofmt/#hdr-The_simplify_command

macat:
Sorry, I misunderstood the PR. You are right, it is not related to Go 1.6.

PR #91, opened 2016-06-12
delete instance

### LINKED ISSUE #86: Delete doesn't clean up instance
Delete method returns with 200 OK, but the instance status is still deleting and some resources are still in kubernetes. No trace of error in logs.

macat:
- Deletes instance from etcd after deleting kubernetes resources

Fixes #86

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -252,28 +252,6 @@ func deploy(s *Server, pID string, ID string) (*instance.Instance, error) {
 	return i, nil
 }
 
-func del(s *Server, pID string, ID string) error {
-	is := services.NewInstanceService(s.store)
-	i, err := is.Show(pID, ID)
-	if err != nil {
-		return nil, err
-	}
-
-	ds := services.NewDeploymentService(s.store, s.playbooks, s.manifests)
-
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -261,10 +261,16 @@ func del(s *Server, pID string, ID string) (*instance.Instance, error) {
 
 	ds := services.NewDeploymentService(s.store, s.playbooks, s.manifests)
 
-	err = ds.DeleteAndNotify(i)
-	if err != nil {
+	if err := ds.DeleteAndNotify(i); err != nil {
```
therealplato: why are both `ds.DeleteAndNotify(i)` and `c.is.Delete(i)` being called? what do they do differently?

macat: `ds.DeleteAndNotify` will delete kubernetes resources while `c.is.Delete` deletes the instance from etcd.

cored:
@macat Are there tests for this? apart from that :+1:

macat:
@cored added check to make sure we are now deleting the instance.

poopoothegorilla:
+1

PR #93, opened 2016-06-23
wait 2 seconds after deleting RCs

macat:
This allows kubernetes to clean up pods after the RC is deleted.

### DIFF
--- a/deployment/manifest_step.go
+++ b/deployment/manifest_step.go
@@ -118,6 +118,7 @@ func (s *ManifestStep) Deploy() error {
 				time.Sleep(200 * time.Millisecond) // Wait for Kubernetes to delete the resource
 				_, err = client.ReplicationControllers(namespace).Get(o.ObjectMeta.Name)
 			}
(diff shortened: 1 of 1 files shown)


deployment/manifest_step.go
```diff
@@ -118,6 +118,7 @@ func (s *ManifestStep) Deploy() error {
 				time.Sleep(200 * time.Millisecond) // Wait for Kubernetes to delete the resource
 				_, err = client.ReplicationControllers(namespace).Get(o.ObjectMeta.Name)
 			}
+			time.Sleep(2 * time.Second) // Wait for Kubernetes to delete pods
```
bobbytables: Doesn't this need to be flipped to get the type to be `time.Duration` ?

bobbytables: Nevermind.

macat: nupe

bobbytables:
👍

PR #96, opened 2016-06-27
Not comparing full object meta objects

macat:


poopoothegorilla:
+1

therealplato:
@macat can you add a description of what behavior this changes?

PR #104, opened 2016-07-12
add quick fix logic

poopoothegorilla:


deployment/manifest_step.go
```diff
@@ -208,11 +208,63 @@ func (s *ManifestStep) Destroy() error {
...
 }
+
+func deleteRC(namespace string, meta *api.ObjectMeta) error {
+	// SCALE RC DOWN TO 0
+	rc, err := client.ReplicationControllers(namespace).Get(meta.Name)
+	if err != nil {
+		return err
+	}
+	var i int32
+	rc.Spec.Replicas = &i
```
therealplato: what's this for?

poopoothegorilla: set replicas to 0

poopoothegorilla: which needs to be *int32 type

deployment/manifest_step.go
```diff
@@ -208,11 +208,63 @@ func (s *ManifestStep) Destroy() error {
...
+	client.ReplicationControllers(namespace).Update(rc)
+	time.Sleep(1 * time.Second) // Wait for Kubernetes to delete pods
+	rc, err = client.ReplicationControllers(namespace).Get(meta.Name)
+	if err != nil {
+		return err
+	}
+	if rc.Status.Replicas != 0 {
+		return errors.New("deployment: RC deletion not successful")
+	}
+	// WATCH REPLICATION CONTROLLER
```
therealplato: will we use this comment in the future?

poopoothegorilla: i think it is better

poopoothegorilla: the sleep is alittle hacking in my opinion but it matches our logic in updating rcs

therealplato:
+1 although i would consider a comment as to why we are using the default int16 pointer value

poopoothegorilla:
@therealplato addressed suggestion

deployment/manifest_step.go
rickyxiv: Lets bump it to 10 seconds this should ensure the pod can terminate cleanly

rickyxiv:
👍

PR #106, opened 2016-07-19
add age to info, sort vars

therealplato:
Adds age since `instance.Created` timestamp to the output of `/bw info`

Be aware, this timestamp gets reset when the instance is *updated* too

This also sorts the `Vars` alphabetically by key

### DIFF
--- a/services/slack.go
+++ b/services/slack.go
@@ -193,9 +193,9 @@ func (c *infoCommand) Execute() (string, error) {
 		glog.Error(msg)
 		return msg, err
 	}
-	vv := Vars{}
+	vv := varSlice{}
 	for k, val := range i.Vars {
-		v := Var{
+		v := varKV{
 			k: k,
 			v: val,
 		}
--- a/services/slack.go
+++ b/services/slack.go
@@ -193,13 +193,22 @@ func (c *infoCommand) Execute() (string, error) {
 		glog.Error(msg)
 		return msg, err
 	}
+	vv := Vars{}
+	for k, val := range i.Vars {
+		v := Var{
+			k: k,
+			v: val,
+		}
+		vv = append(vv, v)
+	}
+	sortVars(vv)
 	m1 := fmt.Sprintf("Playbook: %s\n", wrapQuotes(i.PlaybookID))
 	m2 := fmt.Sprintf("Instance: %s\n", wrapQuotes(i.ID))
 	m3 := fmt.Sprintf("Age: %s\n", wrapQuotes(fmtAge(i.Created)))
 	m4 := fmt.Sprintf("Status: %s\n", wrapQuotes(string(i.Status)))
 	m5 := "Vars:\n"
-	for k, v := range i.Vars {
-		m5 += fmt.Sprintf("  - %s: %s\n", k, wrapQuotes(v))
+	for _, vr := range vv {
+		m5 += fmt.Sprintf("  - %s: %s\n", vr.k, wrapQuotes(vr.v))
 	}
 	msg := m1 + m2 + m3 + m4 + m5
 	return msg, nil
--- a/services/slack.go
+++ b/services/slack.go
@@ -3,6 +3,7 @@ package services
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/namely/broadway/deployment"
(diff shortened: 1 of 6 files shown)


services/sortvars.go
```diff
@@ -0,0 +1,49 @@
...
+// Less returns true if vv[i].k is alphabetically before vv[j].k
+func (vv Vars) Less(i, j int) bool {
+	xx := []Var(vv)
+	if len(xx[i].k) == 0 {
+		return true
+	}
+	if len(xx[j].k) == 0 {
+		return false
+	}
+	return strings.ToLower(xx[i].k)[0] < strings.ToLower(xx[j].k)[0]
```
therealplato: this won't be enough when comparing e.g. `foo=bar; food=baz`

services/slack_test.go
```diff
@@ -256,15 +257,16 @@ func TestInfoExecute(t *testing.T) {
...
 			},
 			"info helloplaybook showinfo",
 			`Playbook: "helloplaybook"
 Instance: "showinfo"
+Age: "3s"
 Status: "deployed"
 Vars:
   - bird: "albatross"
-  - word: ""
+  - word: "phlegmatic"
```
poopoothegorilla: interesting

services/slack_test.go
```diff
@@ -286,6 +288,8 @@ Vars:
 	is := NewInstanceService(store.New())
 	for _, testcase := range testcases {
 		_, err := is.CreateOrUpdate(testcase.Instance)
+		// CreateOrUpdate always resets instance.Created so we can't mock it:
+		time.Sleep(3 * time.Second)
```
poopoothegorilla: why do we sleep before the error?

therealplato: `// CreateOrUpdate always resets instance.Created so we can't mock it:`

i.e. it will set Created to the current timestamp, then the elapsed time will be less than zero. this lets me check that an old instance shows an age value @poopoothegorilla

poopoothegorilla: still don't understand why it would change anything before the err check

poopoothegorilla:
+1

services/slack_test.go
therealplato: thx @poopoothegorilla after f2f you're right that this line is in the wrong place

services/slack.go
```diff
@@ -192,22 +193,45 @@ func (c *infoCommand) Execute() (string, error) {
...
-	msg := m1 + m2 + m3 + m4
+	msg := m1 + m2 + m3 + m4 + m5
 	return msg, nil
 }
 
 func wrapQuotes(s string) string {
 	return fmt.Sprintf("\"%s\"", s)
 }
 
+func fmtAge(d0 int64) string {
```
cored: @therealplato Was playing with something similar yesterday night, this looks good.

services/slack.go
```diff
@@ -192,22 +193,45 @@ func (c *infoCommand) Execute() (string, error) {
 		glog.Error(msg)
 		return msg, err
 	}
+	vv := Vars{}
```
cored: @therealplato What do you think if we extract this into a function in the `instance` package so when we create a new instance we can sort the variables and save it with the sort in it instead of doing it every time that we execute the command from slack?

services/sortvars.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"sort"
+	"strings"
+)
+
+// Implement sort.Interface for Vars:
+
+// Vars is a slice of instance variables
+type Vars []Var
```
cored: :-)

services/sortvars.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Var is a single instance variable with key k, value v
+type Var struct {
```
cored: Isn't this a map ? like `type Var map[string]string` ?

services/sortvars_test.go
```diff
@@ -0,0 +1,50 @@
+package services
```
cored: I'm starting to use this convention for package testing `services_test` instead that way every unexported function won't be available and I'll be just testing public API. What do you think if we do that? If you change your mind on it how can we make the `sortVars()` behavior part of the public interface?

cored:
@therealplato Some comments after that :+1:

services/sortvars.go
therealplato: nope, that's what the deployment package does, but you cannot order maps

bobbytables: Does this need to be exported?

services/sortvars_test.go
bobbytables: @cored just test the package please. There's no gain to using a `_test` package. We all know when a function is exported or not. Lets avoid strict compiler policy and just make our own. Also it can be valuable to test unexported functions sometimes (and honestly, we should do it more for more unit style of tests).

services/slack_test.go
bobbytables: See also: Isaac Rogers

services/slack.go
```diff
@@ -192,22 +193,45 @@ func (c *infoCommand) Execute() (string, error) {
...
+	switch {
+	case age > 24*60*time.Minute:
+		return fmt.Sprintf("%dd", int(age.Hours()/24))
+	case age > 60*time.Minute:
+		return fmt.Sprintf("%dh", int(age.Minutes()/60))
+	case age > 60*time.Second:
+		return fmt.Sprintf("%dm", int(age.Minutes()))
+	}
+	return fmt.Sprintf("%ds", int(age.Seconds()))
+}
```
bobbytables: Soooo.... hahaha. f you convert this to a time.Duration, it has a `String()` func on it that will do this for you.

services/sortvars_test.go
therealplato: at the moment nothing but slack cares about `sortVars` so I think it's correct to remain unexported until something needs it.

I also don't like the idea of only testing public, exported API. If it invokes a private method then you are integration testing without unit testing. Even if you get the integration test right you can still miss the hidden bugs that don't break your public API integration tests, but do break other things

services/slack.go
therealplato: good idea, lets do it in a future PR not here

services/sortvars_test.go
cored: ok

services/slack.go
bobbytables: See: https://golang.org/pkg/time/#Duration

therealplato: Neat!

therealplato: Oh, I saw this, it was too ugly for me :)
```
func (Duration) String

func (d Duration) String() string
String returns a string representing the duration in the form "72h3m0.5s". Leading zero units are omitted. As a special case, durations less than one second format use a smaller unit (milli-, micro-, or nanoseconds) to ensure that the leading digit is non-zero. The zero duration formats as 0, with no unit.
```

services/sortvars.go
therealplato: Nope, good catch, I was going to use it for deployments package but that works fine with `map[string]string` so there is no real gain there

PR #107, opened 2016-07-19
refactor help command and aliases

therealplato:
Old behavior:
- `/bw help` documentation was missing `deploy` *:facepalm:*
- `/bw destroy` did nothing
- No way to list vars

New behavior:
- `/bw destroy...` now works (aliasing `/bw delete`)
- `/bw setvars...` now works (aliasing `/bw setvar`)
- `/bw info repo pr-22`
  - looks up info for playbook `repo` instance `pr-22`
  - replies with a public message to slack containing age and vars
  - all possible vars for this playbook are listed
  - vars are sorted alphabetically
- `/bw help` updated with current commands
- `/bw help` formatting improved


cored:
@therealplato Looks good. What do you think if we just support one way of doing an operation with just one command?

/cc @bobbytables @poopoothegorilla

poopoothegorilla:
@cored what do you mean `we just support one way of doing an operation with just one command?` ... do you mean no aliases?

cored:
@poopoothegorilla yeap; aliases. Why not just support one command for each operation?

therealplato:
> Why not just support one command for each operation?

i forget them all the time and keep trying to use the wrong ones, i've seen others do the same

cored:
@therealplato Wouldn't be better to document the commands instead and point people to the documentation?

therealplato:
imo no, i dont like forcing the user to behave in a particular way, especially when the cost of supporting alternate behaviors is low

cored:
@therealplato cool, then.

poopoothegorilla:
@cored I can see your perspective here but, I think that aliases are really nice to have. It will make for a nicer user experience. 🌈 🐶

poopoothegorilla:
+1

cored:
:+1:

services/slack.go
```diff
@@ -116,9 +116,12 @@ func (c *setvarCommand) playbookContainsVar(playbookID, name string) bool {
...
 
 // CommandHints slack commands help hints
-const commandHints = `/broadway help: This message
-/broadway deploy myPlaybookID myInstanceID: Deploy a new instance
-/broadway setvar myPlaybookID myInstanceID var1=val1 githash=8ad33dad env=prod`
+const commandHints = `
+*/bw deploy myPlaybookID myInstanceID*: Deploy an instance
+*/bw info myPlaybookID myInstanceID*: Display the age and playbook variables of an instance
+*/bw &lt;delete|destroy&gt; myPlaybookID myInstanceID*: Stop and remove an instance
+*/bw &lt;setvar|setvars&gt; myPlaybookID myInstanceID var1=val1 ...* : Set one or more playbook variables for an instance
```
bobbytables: I have a hunch we can make this hint a dynamic,.

services/slack.go
```diff
@@ -116,9 +116,12 @@ func (c *setvarCommand) playbookContainsVar(playbookID, name string) bool {
 }
 
 // CommandHints slack commands help hints
-const commandHints = `/broadway help: This message
-/broadway deploy myPlaybookID myInstanceID: Deploy a new instance
-/broadway setvar myPlaybookID myInstanceID var1=val1 githash=8ad33dad env=prod`
+const commandHints = `
+*/bw deploy myPlaybookID myInstanceID*: Deploy an instance
```
bobbytables: Can we add a code fence to this too actually? That way we'll get monospace font in Slack.

johnnymo87: What is a code fence?

bobbytables: > What is a code fence?

@johnnymo87 - It's the triple backticks that surround things in markdown to give them a "code block" look.

services/slack.go
therealplato: absolutely, I considered it out of scope for this PR though

services/slack.go
therealplato: can we try this as is first? I suspect */bold command*: explanation will be more readable than `/bold command: explanation`

PR #111, opened 2016-07-20
Vendor/cobra+cli

therealplato:
bye godep!
welcome [govendor](https://github.com/kardianos/govendor)!


cored:
:+1:

therealplato:
@macat kubernetes is giving me vendoring indigestion... godep can't find a missing package (from speter.net) and govendor doesn't install all of kubernetes dependencies, see [kubernetes#22427](https://github.com/kubernetes/kubernetes/issues/22427#issuecomment-234124073)

is there any functionality we are using by importing the k8s.io packages, that could not be done through http api against a separate container?

therealplato:
seems that kubernetes just will not work with govendor, dang

PR #112, opened 2016-07-21
Fix for  development environment

cored:
Right now the development environment is not very stable. This is an attempt to fix it.


cored:
We are switching to use minikube #113

PR #113, opened 2016-07-21
Minikube development

bobbytables:


README.md
```diff
@@ -67,6 +67,16 @@ $ kubectl create namespace broadway
...
 ```
 
+### Linux extra setup
+
+In your Virtualbox create a shared folder pointing to where the source code
+currently reside.
+
+```sh
+$ minikube ssh
+(minikuehost)$ sudo mount -t vboxsf -o uid=$(id -u),gid=$(id -g)
```
therealplato: i don't think this will work with a newline and no trailing `\`

Also I recommend using real paths like `sudo mount... /mnt/vbox_sf/foobar /container/mount/point`

README.md
```diff
@@ -67,6 +67,16 @@ $ kubectl create namespace broadway
 $ docker-compose up app
 ```
 
+### Linux extra setup
+
+In your Virtualbox create a shared folder pointing to where the source code
+currently reside.
```
therealplato: "...to where the source code currently resides on your host machine"

docker-compose-dev.yml
```diff
@@ -0,0 +1,23 @@
+version: '2'
```
therealplato: what's the need of the second dockerfile?

README.md
cored: @therealplato what I'm missing is the name of the shared folder on the virtual machine and also the mount point in the host. Cool, this should be updated.

docker-compose-dev.yml
cored: @therealplato since we cannot run minikube on travis or at least I don't know how; we have two different docker compose files one for local development using minikube and the other one for using in travis ci. @bobbytables and I discuss this yesterday.

therealplato: i suggest renames `local-compose.yml` `travis-compose.yml` so it's explicit which to use

cored: ok

bobbytables:
Can we get this finished up soon? We're approaching PR rot. /cc @cored @therealplato

cored:
@bobbytables I'm working on it.

therealplato:
@cored what's the status of this

Issue #102, opened 2016-07-08
Ignore errors while deleting


macat:
Currently delete fails if one of the resources is not available anymore (like a pod). BW should ignore these errors and delete everything it knows of.

PR #114, opened 2016-07-21
Ignore errors

poopoothegorilla:
This will clean up some deletion actions for the RC
fixes https://github.com/namely/broadway/issues/102


deployment/replication_controller.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/v1"
+)
+
+// deployRC deploys the RC after it calls deleteRC
+func deployRC(s *ManifestStep) error {
+	var o *v1.ReplicationController
+	switch s.object.(type) {
+	case *v1.ReplicationController:
```
cored: @poopoothegorilla why do we have two different types ?

deployment/replication_controller.go
```diff
@@ -0,0 +1,113 @@
...
+	client.ReplicationControllers(namespace).Update(rc)
+	time.Sleep(10 * time.Second) // Wait for Kubernetes to delete pods
+	rc, err = client.ReplicationControllers(namespace).Get(metaName)
+	if err != nil {
+		return err
+	}
+	if rc.Status.Replicas != 0 {
+		return errors.New("deployment: RC did not scale successfully")
+	}
+	// WATCH REPLICATION CONTROLLER
```
cored: Can we remove this commented code? @poopoothegorilla

poopoothegorilla: I don't want to remove it yet... since I think we should be switching to this in the future, but for now a simple wait is good

poopoothegorilla: and the wait is how the logic was originally so i don't want to change too much at once

deployment/replication_controller.go
poopoothegorilla: haha yep! it is since `s.object` can be two different versions(k8s api) of a replication controller. It is really confusing. If you figure out a way to only have one type of Replication Controller come in please do update.

cored: @poopoothegorilla just wrap the external collaborator in our own interface. Even if we don't know a proper name you are making decisions based on a type so why don't we introduce our own interface for this?

deployment/replication_controller.go
cored: @poopoothegorilla I would say remove it because we can always use git to recover anything. But is up to you.

therealplato:
+1

deployment/manifest_step.go
```diff
@@ -107,50 +102,18 @@ func compareRCs(a, b *v1.ReplicationController) bool {
 		comparePodSpecs(a.Spec.Template.Spec, b.Spec.Template.Spec)
 }
 
+func comparePods(a, b *v1.Pod) bool {
```
therealplato: this language `compare___` has always confused me. i guess it means `equalPods, equalS, equalI, equalPodSpecs` cc @macat

PR #115, opened 2016-07-22
Refactoring to reduce indirection

cored:
This will be the epic to avoid the need of merging a very big PR.


instance/instance.go
```diff
@@ -2,17 +2,61 @@ package instance
...
 
+// NotFoundError instance not found error
+type NotFoundError string
+
+func (e NotFoundError) Error() string {
+	return fmt.Sprintf("Instance with path: %s was not found", string(e))
+}
+
+// ErrMalformedSaveData in case an instance cannot be marshall
+var ErrMalformedSaveData = errors.New("Saved data for this instance is malformed")
```
poopoothegorilla: i think the package scope should be included here

cored: cool

poopoothegorilla: also lets keep the practice of lowercasing our error messages... https://namely.atlassian.net/wiki/display/CS/Errors

instance/instance.go
```diff
@@ -2,17 +2,61 @@ package instance
...
 
-	"github.com/namely/broadway/env"
+	"github.com/namely/broadway/store"
 )
 
+// NotFoundError instance not found error
+type NotFoundError string
+
+func (e NotFoundError) Error() string {
+	return fmt.Sprintf("Instance with path: %s was not found", string(e))
```
poopoothegorilla: is "with path" needed? or could we just have it `instance: %s not found", string(e)`

instance/instance.go
```diff
@@ -2,17 +2,61 @@ package instance
...
+}
+
+// ErrMalformedSaveData in case an instance cannot be marshall
+var ErrMalformedSaveData = errors.New("Saved data for this instance is malformed")
+
+// Path represents a path for an instance
```
poopoothegorilla: it might be helpful to include an example path "%s/instances/%s/%s" in the doc

instance/instance.go
cored: @poopoothegorilla well, what I'm passing to the error is an Instance path. I don't have any preference at the moment; to me is just more informative to use the entire path.

instance/instance.go
```diff
@@ -2,17 +2,61 @@ package instance
...
+type Path struct {
+	rootPath   string
+	playbookID string
```
poopoothegorilla: what if you use PlaybookPath here?

instance/instance.go
cored: @poopoothegorilla what do you mean?

instance/instance.go
```diff
@@ -2,17 +2,61 @@ package instance
...
+
+// PlaybookPath represents a path for a playbook
+type PlaybookPath struct {
+	rootPath   string
+	playbookID string
+}
+
+// NewPlaybookPath constructor for a new PlaybookPath
+func NewPlaybookPath(rootPath, playbookID string) PlaybookPath {
+	return PlaybookPath{rootPath, playbookID}
```
poopoothegorilla: I was reading this blog post https://medium.com/@cep21/go-client-library-best-practices-83d877d604ca#.ryphk4b6d... the title Directly use the struct kinda makes sense if you aren't setting any special defaults or setups. What are your thoughts?

instance/instance.go
cored: That's a good idea. Will do the change @poopoothegorilla

instance/instance.go
```diff
@@ -41,6 +85,69 @@ func (i *Instance) JSON() (string, error) {
 }
 
 // Path for an instance
-func (i *Instance) Path() string {
-	return env.EtcdPath + "/instances/" + i.PlaybookID + "/" + i.ID
+// func (i *Instance) Path() string {
+// 	return env.EtcdPath + "/instances/" + i.PlaybookID + "/" + i.ID
+// }
```
poopoothegorilla: remove?

instance/instance.go
cored: @poopoothegorilla Yes maybe this is not needed right now due to the lack of construction logic for this type. That's the only reason in which I think is preferable not to use a constructor. Will change it.

instance/instance.go
cored: Cool

instance/instance.go
```diff
@@ -41,6 +85,69 @@ func (i *Instance) JSON() (string, error) {
...
+
+// FindByPath find an instance based on it's path
+func FindByPath(store store.Store, path Path) (*Instance, error) {
+	i := store.Value(path.String())
+	if i == "" {
+		return nil, NotFoundError(path.String())
+	}
+	instance, err := fromJSON(i)
```
poopoothegorilla: if there is an error will instance be `nil`? we could do `return instance, err`? or even `return fromJSON(i)`... but that last one might not be too clear

instance/instance.go
```diff
@@ -41,6 +85,69 @@ func (i *Instance) JSON() (string, error) {
...
+func Save(store store.Store, instance *Instance) error {
+	encoded, err := toJSON(instance)
+	if err != nil {
+		return err
+	}
+	err = store.SetValue(instance.Path.String(), encoded)
+	if err != nil {
+		return err
+	}
+	return nil
```
poopoothegorilla: could we just do `return err`?

instance/instance.go
```diff
@@ -41,6 +85,69 @@ func (i *Instance) JSON() (string, error) {
...
+}
+
+// Delete an instance from the store
+func Delete(store store.Store, path Path) error {
+	return store.Delete(path.String())
```
poopoothegorilla: nice like this

instance/instance.go
```diff
@@ -41,6 +85,69 @@ func (i *Instance) JSON() (string, error) {
...
+}
+
+func fromJSON(jsonData string) (*Instance, error) {
```
poopoothegorilla: I don't quite understand the point of these helper functions... do they really save us any work? I only see them used in 1 place each

instance/instance.go
cored: @poopoothegorilla why do you think is better to return an empty instance?

instance/instance.go
cored: @poopoothegorilla on a successful case? Why?

instance/instance.go
therealplato: to me that implies the whole slash-separated path `/broadway/instances/playbook1`

therealplato: i would consider renaming `type Path -> `type InstancePath`

instance/instance.go
cored: @poopoothegorilla The reason for this was to just not handle errors for something not relating to the operation in hand at the same level of abstraction. The only duplication that those functions remove is the reference to the internal errors but I can remove it if you think it doesn't give any value.

instance/instance_test.go
```diff
@@ -0,0 +1,140 @@
+package instance_test
```
poopoothegorilla: we should discuss this... @therealplato @ehudc @cored 
Do you guys want to do test files like this in the future?

Some projects including the main lib do this to prevent non exported functions from being used in tests.
Keep in mind you have to do what @cored did `. "github.com/namely/broadway/instance"`

cored: @poopoothegorilla what I did was a convenience. In future version I think it would be better to use the `instance.Function` call instead to be explicit about what we are calling. But agree let's discuss it.

instance/instance.go
therealplato: i prefer returning `nil, err`, it just seems cleaner and I see no value in having an empty instance

instance/instance.go
```diff
@@ -41,6 +85,69 @@ func (i *Instance) JSON() (string, error) {
...
+	}
+	instance, err := fromJSON(i)
+	if err != nil {
+		return nil, err
+	}
+	return instance, nil
+}
+
+// FindByPlaybookID find all the instances for an specified playbook path
+func FindByPlaybookID(store store.Store, playbookPath PlaybookPath) ([]*Instance, error) {
```
therealplato: i dont think we should be using a slice of pointers here. i know its old code but still. a slice of instances is already a slice of pointers, this is pointers to pointers

instance/instance_test.go
```diff
@@ -0,0 +1,140 @@
...
+			},
+			ExpectedPlaybookID: "",
+			ExpectedError:      NotFoundError("etcdPath/instances/test/id"),
+		},
+	}
+
+	for _, tc := range testcases {
+		returnedInstance, err := FindByPath(tc.Store, tc.Path)
+		assert.Equal(t, tc.ExpectedError, err, tc.Scenario)
+		if err == nil {
```
poopoothegorilla: what if we change line 56 to `require.Equal`... could that eliminate the need for this check?

instance/instance.go
therealplato: it saves three lines of code, err will be nil

instance/instance.go
cored: @poopoothegorilla do you agree?

instance/instance_test.go
```diff
@@ -0,0 +1,140 @@
...
+			PlaybookPath: NewPlaybookPath("rootPath", "test"),
+			ExpectedInstances: []*Instance{
+				&Instance{PlaybookID: "test", ID: "id", Status: StatusDeployed},
+			},
+			ExpectedError: nil,
+		},
+	}
+	for _, tc := range testcases {
+		_, err := FindByPlaybookID(tc.Store, tc.PlaybookPath)
+		assert.Equal(t, tc.ExpectedError, err, tc.Scenario)
```
poopoothegorilla: should we check to see if the instance id matches? or is that overkill

instance/instance_test.go
therealplato: i think this and the dot import are a bad idea

instance/instance.go
cored: @therealplato Yes; agree with you. How about we leave it for a minor refactor after merging this?

instance/instance_test.go
cored: @poopoothegorilla the intention of that check is not asserting for an error is to not get a nil pointer exception in case an error occurred. I will refactor the test to be more clear of the intention. Thanks for pointing it out.

instance/instance_test.go
```diff
@@ -0,0 +1,140 @@
...
+				},
+			},
+			ExpectedPlaybookID: "",
+			ExpectedError:      NotFoundError("etcdPath/instances/test/id"),
+		},
+	}
+
+	for _, tc := range testcases {
+		returnedInstance, err := FindByPath(tc.Store, tc.Path)
+		assert.Equal(t, tc.ExpectedError, err, tc.Scenario)
```
therealplato: use [`assert.IsType(t, tc.ExpectedError, err, tc.Scenario)`](https://godoc.org/github.com/stretchr/testify/assert#IsType) so it still passes when the error type matches but content is different

instance/instance_test.go
cored: @poopoothegorilla Yes; I think that's a good idea. I don't know why I missed it. I have the expected slice of instances in there.

instance/instance_test.go
```diff
@@ -0,0 +1,140 @@
...
+		PlaybookPath      PlaybookPath
+		ExpectedInstances []*Instance
+		ExpectedError     error
+	}{
+		{
+			Scenario: "When instances exist in the store",
+			Store: &store.FakeStore{
+				MockValues: func(string) map[string]string {
+					return map[string]string{
+						"rootPath/instances/test": `{"playbook_id": "test", "id": "id", "status": "deployed"}`,
```
therealplato: did you change the paths?
i thought this would be `rootPath/instances/test/id`?

instance/instance_test.go
cored: @therealplato Well if you take a look at the `ExpectedError` definition is specific to the defined type. So the `assert.IsType` is implicit. This function will always return a predefined type so, I think this is save to do. What do you think?

store/et/etcd.go
```diff
@@ -1,4 +1,4 @@
-package store
+package et
```
therealplato: this package name does not describe what it does. what's wrong with `etcd`

services/deployment.go
```diff
@@ -15,15 +15,15 @@ import (
...
+	store     store.Store
 	playbooks map[string]*deployment.Playbook
 	manifests map[string]*deployment.Manifest
 }
 
 // NewDeploymentService creates a new DeploymentService
 func NewDeploymentService(s store.Store, ps map[string]*deployment.Playbook, ms map[string]*deployment.Manifest) *DeploymentService {
 	return &DeploymentService{
-		repo:      instance.NewRepo(s),
+		store:     s,
```
therealplato: 👍

services/instance.go
```diff
@@ -157,17 +162,19 @@ func (is *InstanceService) Show(playbookID, ID string) (*instance.Instance, erro
...
 func (is *InstanceService) AllWithPlaybookID(playbookID string) ([]*instance.Instance, error) {
-	return is.repo.FindByPlaybookID(playbookID)
+	playbookPath := instance.NewPlaybookPath(env.EtcdPath, playbookID)
+	return instance.FindByPlaybookID(is.store, playbookPath)
 }
 
 // Delete removes an instance
 func (is *InstanceService) Delete(i *instance.Instance) error {
-	ii, err := is.Show(i.PlaybookID, i.ID)
+	_, err := is.Show(i.PlaybookID, i.ID)
```
therealplato: do we care about erroring when you try to delete a missing or malformed instance?
should we just skip this check and try to delete it no matter what?
will `instance.Delete()` error?

services/slack_test.go
```diff
@@ -291,17 +291,17 @@ Vars:
 			&instance.Instance{PlaybookID: "helloplaybook", ID: "randomid"},
 			"info helloplaybook showmissing",
 			"Failed to retrieve info for helloplaybook/showmissing: Instance not found",
-			instance.NotFound{},
+			instance.NotFoundError("instances//"),
```
therealplato: what is the `//`?

services/deployment.go
```diff
@@ -15,15 +15,15 @@ import (
 
 // DeploymentService implements the Broadway logic for deployments
 type DeploymentService struct {
-	repo      instance.Repository
+	store     store.Store
```
poopoothegorilla: NICE

instance/instance.go
therealplato: fine by me but please add a TODO flag so we remember later

instance/instance_test.go
therealplato: ok, its fine. i like following this pattern always, but it is only necessary when the error message is not something you can mock

instance/instance_test.go
therealplato: if this is a playbook path, why does it have `"status": "deployed"`

instance/instance.go
poopoothegorilla: okay i am good with that... its more explicit

instance/instance.go
poopoothegorilla: yeh just a thought

instance/instance_test.go
poopoothegorilla: So i don't think the dot imports are a bad idea. It doesn't/wouldn't break anything... but i think for our use cases it isn't really needed. They seem to be beneficial for packages that are being used by many applications(like the standard lib). My vote is lets not use this for broadway

@cored any strong opinions on keeping this?

instance/instance_test.go
poopoothegorilla: oh good point! sorry misread that :P it is fine

instance/instance.go
cored: @poopoothegorilla Sure; I'll lower the case down. I was reading the documentation for pq and they actually are signaling that the error comes from the pq library not package. So in that case the error should be `broadway: <string>`. 
https://github.com/lib/pq/blob/master/notify_test.go#L24

instance/instance.go
cored: @therealplato shouldn't that be kinda redundant, taking into account that the `Path` is scope in the `instance` namespace ?

instance/instance.go
cored: @therealplato @poopoothegorilla Wouldn't that be less clear ? Like we won't be explicit about showing what is the state of err to me it's reads like 'store will always return an error' instead 'we check if the errors exist and do something with it'. What do you guys think?

instance/instance_test.go
cored: @poopoothegorilla I only use the dot notation for testing; I always try to be explicit about the exported function I'm using. 
@therealplato In my experience if you have a private identifier that needs to be tested maybe you are missing an abstraction that needs to be public or part of another component. Seems this API is so straight forward I didn't see the need to test the internals of the package at the moment seems those functions get stressed by the test. In terms of convinience I agree with you if you want to test private stuffs then go ahead and use `package` without the `_test`.

instance/instance.go
cored: @therealplato I didn't the get very clear the part in which an slice of instance is already an slice of pointers. Are you saying that `[]*Instance` is the same as `[]Instance` ? How is that possible, if that's the case; if not could you elaborate?

instance/instance_test.go
cored: @therealplato what do you mean by `the error message is not something you can mock`? Can you write an example code?

instance/instance_test.go
cored: @therealplato so what we are actually saving into etcd is a JSON "playbookPath": ["instances"]" what contains the status is the data for the instance. Does that make sense?

instance/instance.go
therealplato: @cored you're right. jim's comment confused me :)

store/et/etcd.go
cored: @therealplato naming collision. If you have a better idea I'm open to it; I also don't like the name.

services/instance.go
cored: @therealplato Will have to check the etcd test to know if a delete will fail with malformed data or missing data. Also take into account that probably this function will go since the services layer will be part of the same namespace `instance`. Good catch though.

instance/instance.go
therealplato: good question, i don't know yet! im gonna read this https://blog.golang.org/go-slices-usage-and-internals

services/slack_test.go
cored: @therealplato That's the path for the instance. Since still there's some code missing to build the proper `Path` type to an instance the `String()` method of that type will return that format since the zero value for an string is empty string that's why looks like this.

instance/instance.go
therealplato: but we don't do something with it - we just return it up. i think jim's suggestion is fine, for the case where we immediately return in both code paths

cored: @therealplato I mean, what I'm saying is from the point of view of the reader. If you read both versions in your head is kinda confusing because is saying that is always returning an error. So you will have to make the condition in your head creating more overhead, don't you think? I can switch it up if you guys think is clear.

therealplato: i see your point, the compiler will probably end up performing this optimization anyway

I personally do find `return store.SetValue(instance.Path.String(), encoded)` to be clear, especially with a 1-ary return signature

instance/instance_test.go
therealplato: here's a contrived example of a dynamic error https://play.golang.org/p/-eIrhGVlfc we ran into this while testing unlocks

instance/instance_test.go
therealplato: I get that the playbook path corresponds to an etcd tree node that has instances as child leaf nodes.

What I don't get is why you are using this specific data for the mock store.

Why not use:

```
map[string]string{
	"rootPath/instances/test/id": `{"playbook_id": "test", "id": "id", "status": "deployed"}`,
	"rootPath/instances/test/id2": `{"playbook_id": "test", "id": "id2", "status": "deploying"}`,
}
```

instance/instance_test.go
cored: @therealplato Well that case won't apply to my test; I'm actually returning the type and specific type.

instance/instance_test.go
cored: @therealplato Well, kinda the same but I will add more if you think is a good idea.

instance/instance_test.go
therealplato: yep, keep what you have, as we discussed f2f my case above is a general case where you can handle dynamic errors with assert.IsType

store/et/etcd.go
cored: @therealplato namespace collision; we are using core os etcd which uses that namespace.

instance/instance.go
bobbytables: You would typically do `broadway/instance: something broke`

instance/instance.go
bobbytables: I think this is fine right now TBH. I like having explicit constructors in Go.

instance/instance.go
bobbytables: Holy bikeshedding batman.

instance/instance.go
cored: @bobbytables I have a PR open to remove the constructors for now. #117

instance/instance.go
bobbytables: You should really just `return err` if that's the only thing returned for a function. There's no gain to having the `if` statement.

cored: ok

instance/instance.go
bobbytables: I think this is fine.

instance/instance_test.go
bobbytables: I'm really not a fan of `_test` packages. They give you no gain at all and actually just become more cumbersome. They also add more mental overhead. Go handles testing very well, lets use the same package please.

cored: Sure

instance/repository.go
```diff
@@ -1,100 +0,0 @@
...
-		instances = append(instances, &instance)
-	}
-
-	return instances, nil
-}
-
-// Delete removes an instance
-func (ir *Repo) Delete(i *Instance) error {
-	return ir.store.Delete(i.Path())
-}
```
bobbytables: Mmmmmm 🔻

store/et/etcd.go
bobbytables: You can just import the package you want. This really should be `etcd` and use imports like:

```go
import bwetcd "github.com/namely/broadway/store/etcd"
```

cored: Will change it.

poopoothegorilla:
+1

therealplato:
+1 gj

PR #119, opened 2016-07-28
Feature/locking instances

### LINKED ISSUE #99: Ability to lock an instance
It would be super convenient to lock an instance. This could be good for moratoriums, qa cycles, etc.

It could operate like:

```sh
/bw lock {playbook} {instance}
```

Any attempt to deploy a locked instance would yield a response in Slack:

```sh
/bw deploy {playbook} {instance}

Broadway: "The instance is currently locked by @bobbytables"
```

```sh
/bw destroy {playbook} {instance}

Broadway: "The instance is currently locked by @bobbytables"
```

cored:
Fix #99


instance/instance.go
```diff
@@ -117,6 +123,20 @@ func Delete(store store.Store, path Path) error {
...
 
+// Lock an instance
+func Lock(store store.Store, path Path) (*Instance, error) {
+	instance, err := FindByPath(store, path)
+	if err != nil {
+		return nil, err
+	}
+	instance.Status = StatusLocked
+	err = Save(store, instance)
+	if err != nil {
```
poopoothegorilla: scope

instance/instance.go
```diff
@@ -117,6 +123,20 @@ func Delete(store store.Store, path Path) error {
 	return store.Delete(path.String())
 }
 
+// Lock an instance
+func Lock(store store.Store, path Path) (*Instance, error) {
```
poopoothegorilla: Any reason you chose to find an instance by path vs just having the method `Lock` on `Instance`?

instance/instance.go
cored: @poopoothegorilla ?

instance/instance.go
cored: @poopoothegorilla you mean to do something like `instance.Lock()` ? If that's the case what would be the benefit of that?

poopoothegorilla: it is fine anyway though

services/slack.go
```diff
@@ -22,6 +23,20 @@ type deployCommand struct {
...
 }
 
+type lockCommand struct {
+	pID string
+	ID  string
+}
+
+func (c *lockCommand) Execute() (string, error) {
+	path := instance.Path{env.EtcdPath, c.pID, c.ID}
+	i, err := instance.Lock(etcdstore.New(), path)
```
poopoothegorilla: nice

services/slack_test.go
```diff
@@ -321,3 +336,47 @@ Vars:
...
+		}
+		command := BuildSlackCommand(
+			testcase.Args,
+			is,
+			map[string]*deployment.Playbook{
+				"helloplaybook": {ID: "showinfo"},
+			},
+		)
+		// CreateOrUpdate always resets instance.Created so we can't mock it:
+		time.Sleep(3 * time.Second)
```
poopoothegorilla: why do we need this?

services/slack_test.go
```diff
@@ -321,3 +336,47 @@ Vars:
...
+			"/broadwaytest/instances/helloplaybook/showinfo is currently locked",
+			nil,
+		},
+	}
+
+	is := NewInstanceService(etcdstore.New())
+	for _, testcase := range testcases {
+		_, err := is.CreateOrUpdate(testcase.Instance)
+		if err != nil {
+			t.Log(err)
```
poopoothegorilla: was this debugging t.Log?

cored: @poopoothegorilla yes; well since we are not testing the create instance service we just want to know what happen in case of a creation failure.

instance/instance.go
poopoothegorilla: `if err = Save(store, instance); err != nil {`

instance/instance.go
poopoothegorilla: @cored that isn't what i meant... but no worries i think this way works.

services/slack_test.go
poopoothegorilla: k

poopoothegorilla:
when all pass +1

services/slack_test.go
```diff
@@ -321,3 +336,47 @@ Vars:
...
+		Args        string
+		ExpectedMsg string
+		ExpectedErr error
+	}{
+		{
+			"When passing locking command to an instance",
+			&instance.Instance{
+
+				PlaybookID: "helloplaybook",
+				ID:         "showinfo",
```
therealplato: can you rename this to `locktest` so we have some useful information when tests fail

services/slack_test.go
therealplato: we shouldn't need this, this was specifically for testing the age of instances returned right

services/slack_test.go
cored: ok

services/slack_test.go
cored: I'll remove it.

instance/instance.go
therealplato: @cored why did you skip this one

cored: @therealplato maybe I missed it; I'll change it the unlock feature.

therealplato:
bobby and I noticed that this probably won't work until we fix https://github.com/namely/broadway-nginx/blob/master/templates/nginx.conf.tmpl#L9 and line 37

PR #121, opened 2016-07-30
Refactor-init/base

therealplato:
- add a cli
- inject configuration as a dependency from the top down
- refactors of other broadway packages are PR'd onto this base PR for clarity's sake:
  - [vendor (therealplato#1)](https://github.com/therealplato/broadway/pull/1)
  - [server (therealplato#2)](https://github.com/therealplato/broadway/pull/2)
  - [deployment (therealplato#3)](https://github.com/therealplato/broadway/pull/3)
  - [services (therealplato#4)](https://github.com/therealplato/broadway/pull/4)
  - [store+notifications (therealplato#5)](https://github.com/therealplato/broadway/pull/5)

---

my first rebase brought 33 commits down to ~25  
then a second rebase pulled out these 12 that form the base functionality of `cfg` and `cmd` packages  
a third, fourth and fifth were used to fix the second, which accidentally included vendor/ and other packages...


env/env.go
```diff
@@ -1,4 +1,4 @@
-package env
+package notenv
 
 import (
 	"os"
```
bobbytables: lets just remove the whole package now :)

cfg/cfg.go
```diff
@@ -0,0 +1,24 @@
+package cfg
+
+// GlobalCfg is this deployment's global configuration object
+// Only touch this in main.go and cmd/*.go, inject cfg dependency into all other code
+var GlobalCfg Type
+
+// Type declares what the common config looks like
+type Type struct {
```
cored: why type and not `Config`?

therealplato: i first had it as `cfg.CommonCfgType` then `cfg.CfgType` and govet complained that it stuttered so I went to the simplest case `cfg.Type`

poopoothegorilla:
+1 for all dependent prs

cored:
@therealplato I left you some comments; after that +1

poopoothegorilla:
finally +1

PR #122, opened 2016-08-01
CORE-395: Unlock an instance

cored:


instance/instance_test.go
```diff
@@ -242,3 +242,90 @@ func TestLock(t *testing.T) {
...
+			ExpectedInstance: nil,
+		},
+	}
+
+	for _, tc := range testcases {
+		instance, err := Unlock(tc.Store, tc.Path)
+		assert.Equal(t, tc.ExpectedError, err, tc.Scenario)
+		assert.Equal(t, tc.ExpectedInstance, instance, tc.Scenario)
+	}
+}
```
bobbytables: This is nice.

services/slack.go
```diff
@@ -17,10 +17,18 @@ type SlackCommand interface {
...
-	is  *InstanceService
+}
+
+func (c *unlockCommand) Execute() (string, error) {
+	path := instance.Path{env.EtcdPath, c.pID, c.ID}
+	i, err := instance.Unlock(etcdstore.New(), path)
+	if err != nil {
+		return "", err
+	}
+	return i.String(), nil
```
bobbytables: SO simple. 🔥

bobbytables:
👍

instance/instance.go
```diff
@@ -8,6 +8,13 @@ import (
 	"github.com/namely/broadway/store"
 )
 
+// NotLockedStatusError instance is not locked
+type NotLockedStatusError string
+
+func (e NotLockedStatusError) Error() string {
+	return fmt.Sprintf("broadway/instance: %s is not locked", string(e))
```
poopoothegorilla: why did you choose to add the `broadway/` infront?

cored: @poopoothegorilla the reason why `pq` package which was your source of inspiration to be specific about error types uses `pq` in the front is becaues that's the name of the library+package. We are using `broadway` as the name of the library and `instance` is the name of the package.

therealplato:
+1 although i think we will need to refactor this in a future PR to have both `instance.Locked` and `instance.Status` - how else will you know whether it becomes `created` or `deployed` when you unlock it?

cored:
@therealplato I don't get it? Isn't the status the representation of that?

therealplato:
> Isn't the status the representation of that?

yes, but you lost that representation when you overwrite `i.Status=StatusLocked`,

cored:
@therealplato Yes; I will fix this.

instance/instance.go
```diff
@@ -137,6 +146,23 @@ func Lock(store store.Store, path Path) (*Instance, error) {
...
+	instance, err := FindByPath(store, path)
+	if err != nil {
+		return nil, err
+	}
+	if instance.Status != StatusLocked {
+		return nil, NotLockedStatusError(path.String())
+	}
+	instance.Status = StatusUnlocked
+	err = Save(store, instance)
+	if err != nil {
```
poopoothegorilla: scope

services/slack_test.go
```diff
@@ -375,7 +375,50 @@ func TestLockExecute(t *testing.T) {
...
+			},
+			"unlock helloplaybook locktest",
+			"/broadwaytest/instances/helloplaybook/locktest is currently unlocked",
+			nil,
+		},
+	}
+
+	is := NewInstanceService(etcdstore.New())
+	for _, testcase := range testcases {
+		_, err := is.CreateOrUpdate(testcase.Instance)
```
poopoothegorilla: what if there is an error creating or updating? should the test continue?

poopoothegorilla:
+1

services/slack_test.go
cored: @poopoothegorilla if something fails it will log the error and also fail so it doesn't matter. Do you have any suggestion on this?

poopoothegorilla: okay thats fine I guess I was just thinking that if it failed why do the rest of the tests if they will fail as well?

cored:
@therealplato I reflected what we discussed yesterday. Added a `Lock` field to the `Instance` struct now we are going to save the information related to status. Too

therealplato:
cool +1

PR #123, opened 2016-08-01
Fix locking state

cored:
This will fix the locking state so we don't loose information regarding the status of the instance when locking it. 

cc @therealplato @bobbytables


therealplato:
+1

PR #126, opened 2016-08-05
Manifest package

cored:
The reason for this refactoring is to reduce the amount of integration test that we have right now. The test suite takes too long locally to run.


manifest/manifest.go
```diff
@@ -0,0 +1,73 @@
...
+	"path/filepath"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// ManifestExtension is added to each task manifest item to make a filename
+var ManifestExtension = ".yml"
```
therealplato: use my new ConfigType.ManifestExtension instead of hardcoding this please

cored: @therealplato Oh, I just copy and paste the files after your refactor; maybe you missed it? I'll do it, sure.

therealplato: yeah, whoops!

i really would prefer to have all business logic in services/manifest.go and only have the manifest file be a dumb type

cored: @therealplato I'm following another approach but similar to that idea. Take a look at the `instance` package. It have the definition of the data in there and also the operations that you can do on top of those structs. The idea is to be specific about the operations that work on top of a certain data. In my mind I think that at some point we are going to remove the `services` package I think it was a premature abstraction on my part to think in that package.

Issue #80, opened 2016-05-20
Add TTL + automatic deletion of instances


macat:
This feature will enable maintainers to set TTL for instances, than at the expiration time resources will be freed up.

- [ ] Add `ttl` (timestamp - int64) to Instance
- [ ] Periodically query all instances
- [ ] Filter results by ttl<now
- [ ] Attempt to delete each candidate
- [ ] Notify slack channel

PR #127, opened 2016-08-08
CORE-385:   Add TTL + automatic deletion of instances

cored:
Fix: #80


instance/instance.go
```diff
@@ -4,10 +4,17 @@ import (
...
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/namely/broadway/store"
 )
 
+// NewExpiredAt builds a new ExpiredAt
+func NewExpiredAt(daysToExpire int, from time.Time) time.Time {
+	from.AddDate(from.Year(), int(from.Month()), from.Day()+daysToExpire)
```
therealplato: what happens when from.Day is 29 and you add 5 days to expire?
how about `return from.Add(time.Day * time.Duration(daysToExpire))`

instance/instance.go
```diff
@@ -177,3 +187,17 @@ func toJSON(instance *Instance) (string, error) {
 	}
 	return string(encoded), nil
 }
+
+func retrieveInstancesByKey(store store.Store, key string) ([]*Instance, error) {
+	data := store.Values(key)
+	instances := []*Instance{}
```
therealplato: can you use `[]Instance` here?

instance/instance.go
```diff
@@ -25,6 +32,13 @@ func (e NotFoundError) Error() string {
 // ErrMalformedSaveData in case an instance cannot be marshall
 var ErrMalformedSaveData = errors.New("broadway/instance: saved data for this instance is malformed")
 
+// GlobalPath represents the global path key to save an instance in store
+type GlobalPath string
+
+func (gp GlobalPath) String() string {
+	return fmt.Sprintf("%s/instances", string(gp))
+}
+
```
therealplato: i dont think this is any better than `var GlobalPath = PathPrefix + "/instances"`

services/instance.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/namely/broadway/store"
 )
 
+// InstanceExpirationDay represents the amount of days to expire an instance
+const InstanceExpirationDay = 5
```
therealplato: use `InstanceExpirationDays`, this could be misinterpreted as "thursday" or "friday"

instance/instance.go
cored: @therealplato when the day is 29 it's properly rotated to the next month. I added a new unit test so you can see it working. 
In another note; I didn't see that constant in the time package documentation `Day`. 
https://golang.org/pkg/time/
There's been discussion about adding it but it doesn't exist. 
https://github.com/golang/go/issues/11473

instance/instance.go
cored: @therealplato why do you think that will be better?

instance/instance.go
cored: @therealplato Any reason for going on that route?

services/instance.go
cored: sure

instance/instance.go
therealplato: got it, thx

instance/instance.go
therealplato: as usual I dont like adding a new type that does not have behavior attached, and at the moment there is no reason to set GlobalPath dynamically

instance/instance.go
therealplato: slices are already pointers to their elements, i think this makes it hold pointers to pointers

instance/instance.go
cored: @therealplato the code in the service is using this type right now; I can remove it an pass an string if you prefer it like that; however I prefer to change it in another PR.

instance/instance.go
cored: in what way does that create an problem? @therealplato I still can't see the issue.

therealplato: `[]*Instance` is harder to comprehend and more complicated to use than `[]Instance`

cored: @therealplato do you mind if we ask for a second opinion? I still can't see what is the complication. @bobbytables @macat @poopoothegorilla

instance/instance.go
bobbytables: A day is a variable unit of time so it probably won't change in Go.

instance/instance.go
bobbytables: I agree with Isaac, this is an over abstraction.

instance/instance.go
```diff
@@ -105,17 +111,12 @@ func FindByPath(store store.Store, path Path) (*Instance, error) {
...
-		}
-		instances = append(instances, instance)
-	}
+	return retrieveInstancesByKey(store, playbookPath.String())
+}
 
-	return instances, nil
+// All find all instances from in the store
+func All(store store.Store, globalPath GlobalPath) ([]*Instance, error) {
+	return retrieveInstancesByKey(store, globalPath.String())
```
bobbytables: Yeah like the GlobalPath in here is just adding more to the function signature. It's not necessary.

instance/instance.go
cored: @bobbytables Sure; I'll remove this abstraction.

server/server.go
```diff
@@ -74,6 +79,14 @@ func (s *Server) Init() {
 
 	s.playbooks = deployment.AllPlaybooks
 	glog.Infof("Server Playbooks: %+v", s.playbooks)
+
+	glog.Info("Initialize deployed instances cleanup worker")
+	ds := services.NewDeploymentService(s.Cfg, s.store, s.playbooks, s.manifests)
+	go func() {
+		for {
+			t = timer.AfterFunc(time.Hour*24*time.Duration(daysForInstanceCleanup), ds.RemoveExpiredInstances())
```
poopoothegorilla: sweet

instance/instance.go
bobbytables: You should really only use a pointer when you need to modify the object when iterating over it. I'm of the fact that we're not going to save valuable memory here by using a pointer. Go uses pointers to show intent as well, and using a pointer here may give mixed signals.

services/deployment_test.go
```diff
@@ -17,6 +18,56 @@ func init() {
...
+	expirationDate := time.Date(2016, 8, 05, 16, 00, 00, 651387237, time.UTC)
+	expiredAt := instance.NewExpiredAt(InstanceExpirationDays, currentTime)
+
+	cases := []struct {
+		Scenario string
+		Instance *instance.Instance
+		Error    error
+	}{
+		{
+			Scenario: "RemoveExpiredInstances: When an instance just expired",
```
poopoothegorilla: are there more scenarios?

instance/instance.go
cored: @bobbytables cool, will remove the pointer declaration.

server/server.go
bobbytables: Why not do every 15 seconds?

server/server.go
```diff
@@ -28,9 +29,13 @@ type Server struct {
 	Cfg        cfg.Type
 }
 
-const commandHint string = `/broadway help: This message
+const (
+	commandHint string = `/broadway help: This message
 /broadway deploy myPlaybookID myInstanceID: Deploy a new instance`
 
+	daysForInstanceCleanup = 5
```
poopoothegorilla: I am guessing this will go away later?

services/deployment.go
```diff
@@ -220,6 +221,26 @@ func (d *DeploymentService) DeleteAndNotify(i *instance.Instance) error {
...
+// RemoveExpiredInstances remove expired instances from the deployment
+func (d *DeploymentService) RemoveExpiredInstances(expirationDate time.Time) error {
+	globalPath := fmt.Sprintf("%s/instances", d.Cfg.EtcdPath)
+	instances, err := instance.All(d.store, globalPath)
+	if err != nil {
+		return err
+	}
+	for _, i := range instances {
+		if i.Status == instance.StatusDeployed {
+			if i.ExpiredAt < expirationDate.Unix() {
```
bobbytables: nitty gritty but this should be `<=` I think. If it matches exactly it should be considered expired.

server/server.go
cored: Did not have any time specified so I picked up one. @bobbytables will add the 15 seconds interval.

services/instance.go
```diff
@@ -181,6 +189,9 @@ func (is *InstanceService) Delete(i *instance.Instance) error {
 	return instance.Delete(is.store, path)
 }
 
+// func (is *InstanceService) RemoveExpiredInstances() error {
+// }
```
bobbytables: Ditch?

server/server.go
poopoothegorilla: or what if you used `instance.InstanceExpirationDays` instead?

services/instance.go
```diff
@@ -181,6 +189,9 @@ func (is *InstanceService) Delete(i *instance.Instance) error {
 	return instance.Delete(is.store, path)
 }
 
+// func (is *InstanceService) RemoveExpiredInstances() error {
```
poopoothegorilla: do we still want this?

cored: @poopoothegorilla good catch, thanks

services/deployment.go
cored: @bobbytables cool

store/etcdstore/etcd.go
```diff
@@ -60,9 +61,12 @@ func (*etcdStore) Values(path string) (values map[string]string) {
 		glog.Warningf("Ignoring error getting values: %s. Error: %s", path, err.Error())
 		return values
 	}
+	fmt.Println(resp)
```
poopoothegorilla: needed still?

server/server.go
cored: @poopoothegorilla one variable is to specified the amount of time to create the expiration date the other one is to specified how often we do the cleanup.

store/etcdstore/etcd.go
```diff
@@ -60,9 +61,12 @@ func (*etcdStore) Values(path string) (values map[string]string) {
...
 	if resp.Node != nil && len(resp.Node.Nodes) > 0 {
 		for _, node := range resp.Node.Nodes {
-			values[lastKeyItem(node.Key)] = node.Value
+			if node.Dir == false {
```
poopoothegorilla: what is this?

server/server.go
poopoothegorilla: oh good point nice thanks for the clarification

instance/instance.go
cored: @therealplato @bobbytables if you guys checkout the code in `toJSON` function which is the function doing the marshalling is returning a pointer and that was the reason for me to use an slice of instance pointers so I can use append without extracting the value from the pointer. What about if we do this change later, I mean if there's a very specific benefit of changing this right now I can start the refactoring of the other functions in the `instance` package to use a value instead of a pointer. Let me know your opinions.

store/etcdstore/etcd.go
```diff
@@ -62,7 +62,9 @@ func (*etcdStore) Values(path string) (values map[string]string) {
 	}
 	if resp.Node != nil && len(resp.Node.Nodes) > 0 {
 		for _, node := range resp.Node.Nodes {
-			values[lastKeyItem(node.Key)] = node.Value
+			if node.Dir == false {
```
poopoothegorilla: @cored could you explain what this is doing?

cored: @poopoothegorilla a node in etcd could be a directory or an actual node. A directory contains several nodes with values but a directory does not contain a value. So we need to make the distinction. I'm writing a test right now to stress this out it's been stressed indirectly by the integration test on the deployment service.

services/deployment_test.go
cored: @poopoothegorilla There's a new one for removing if the expiration date is the same as current date.

store/etcdstore/etcd.go
poopoothegorilla: cool just ping me when you got those :)

poopoothegorilla:
when tests pass +1

services/deployment.go
```diff
@@ -220,6 +221,26 @@ func (d *DeploymentService) DeleteAndNotify(i *instance.Instance) error {
 
 }
 
+// RemoveExpiredInstances remove expired instances from the deployment
+func (d *DeploymentService) RemoveExpiredInstances(expirationDate time.Time) error {
+	globalPath := fmt.Sprintf("%s/instances", d.Cfg.EtcdPath)
```
macat: Can you add logging here. I would like to see when this method runs and also some info about how many instances it found.

server/server.go
```diff
@@ -74,6 +72,14 @@ func (s *Server) Init() {
 
 	s.playbooks = deployment.AllPlaybooks
 	glog.Infof("Server Playbooks: %+v", s.playbooks)
+
+	glog.Info("Initialize deployed instances cleanup worker")
+	ds := services.NewDeploymentService(s.Cfg, s.store, s.playbooks, s.manifests)
+	go func() {
+		for {
+			t = timer.AfterFunc(time.Hour*12*time.Duration(time.Hour), ds.RemoveExpiredInstances())
```
macat: This needs to be configurable from env vars.

instance/instance.go
therealplato: this actually shouldn't work because you're not assigning `from` before you return it
i still suggest `return from.Add(time.Hour * 24 * time.Duration(daysToExpire))`

instance/instance.go
```diff
@@ -72,24 +80,15 @@ const (
 	// StatusNew represents a newly created instance
 	StatusNew Status = ""
 	// StatusDeploying represents an instance that has begun deployment
-	StatusDeploying = "deploying"
+	StatusDeploying Status = "deploying"
```
therealplato: why did you add these?

instance/instance.go
```diff
@@ -105,17 +104,12 @@ func FindByPath(store store.Store, path Path) (*Instance, error) {
 
 // FindByPlaybookID find all the instances for an specified playbook path
```
therealplato: let's rename this if it's no longer taking just the playbook ID

instance/instance.go
```diff
@@ -177,3 +171,17 @@ func toJSON(instance *Instance) (string, error) {
 	}
 	return string(encoded), nil
 }
+
+func retrieveInstancesByKey(store store.Store, key string) ([]*Instance, error) {
```
therealplato: we can move this behavior into All and remove this function

instance/instance.go
cored: @therealplato an string is not the same as an `Status` when I was asserting on the status the type check failed.

instance/instance.go
cored: @therealplato ? I don't get what you mean.

instance/instance.go
cored: ok

instance/instance_test.go
```diff
@@ -329,3 +369,35 @@ func TestUnlock(t *testing.T) {
...
+		ExpectedInstances []*Instance
+		ExpectedError     error
+	}{
+		{
+			Scenario: "All: When the instance is properly save",
+			Path:     "broadwaytest/instances",
+			Store: &store.FakeStore{
+				MockValues: func(path string) map[string]string {
+					return map[string]string{
+						"etcdPath/instances": `{"playbook_id":"test", "id": "id", "status": "deployed"}`,
```
therealplato: can you return more than one instance here? in real life usage we will usually have more than one and i'd like the test to reflect that scenario

server/server.go
```diff
@@ -28,9 +29,6 @@ type Server struct {
 	Cfg        cfg.Type
 }
 
-const commandHint string = `/broadway help: This message
-/broadway deploy myPlaybookID myInstanceID: Deploy a new instance`
-
```
therealplato: thx

services/deployment.go
```diff
@@ -220,6 +221,26 @@ func (d *DeploymentService) DeleteAndNotify(i *instance.Instance) error {
...
+	instances, err := instance.All(d.store, globalPath)
+	if err != nil {
+		return err
+	}
+	for _, i := range instances {
+		if i.Status == instance.StatusDeployed {
```
therealplato: should we remove deployed but locked instances @bobbytables @macat

instance/instance_test.go
cored: @therealplato there is a test reflecting that in the `etcd_test`. What will be the benefit of also adding it in here?

services/deployment.go
```diff
@@ -220,6 +221,26 @@ func (d *DeploymentService) DeleteAndNotify(i *instance.Instance) error {
...
+			if i.ExpiredAt <= expirationDate.Unix() {
```
therealplato: let's cache `x := expirationDate.Unix()` outside this loop

services/instance.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/namely/broadway/store"
 )
 
+// InstanceExpirationDays represents the amount of days to expire an instance
+const InstanceExpirationDays = 5
```
therealplato: can you add `TODO: inject via config objects`

instance/instance.go
therealplato: @macat we were wrong about this behavior :o

instance/instance.go
therealplato: gorename to "FindByPlaybookPath" because the path is not the id

instance/instance_test.go
therealplato: you would be testing "the All function successfully converts multiple etcd json into instances", the previous test is only testing retrieving json

cored: I will remove this function either way

instance/instance.go
macat: ups, sorry.

macat:
+1

PR #134, opened 2016-08-18
Simplify Broadway by removing pod manifests

macat:
- remove tasks and pod manifests from playbooks


poopoothegorilla:
+1

pkg/deployment/playbook_test.go
```diff
@@ -25,44 +25,22 @@ vars:
...
+name: The Project
 messages:
   created: "help I'm stuck in a test {{ .ID }}"
   deployed: "help I'm still stuck in a test {{ .ID"
-tasks:
-  - name: Deploy Postgres
-    manifests:
-      - hello
+manifests:
+  - hellooo
```
therealplato: let's use `hello` here so there's no ambiguity about whether the failure is the bad template or the missing manifest

pkg/server/server_test.go
```diff
@@ -22,13 +22,15 @@ import (
...
 var testCfg = cfg.Type{
+	K8sNamespace:       "broadway",
 	AuthBearerToken:    "testtoken",
 	SlackToken:         testToken,
 	ManifestsPath:      "../../examples/manifests",
 	ManifestsExtension: ".yml",
 	PlaybooksPath:      "../../examples/playbooks",
 	EtcdEndpoints:      "http://localhost:4001",
 	EtcdPath:           "/broadwaytest",
+	InstanceCleanup:    15,
```
therealplato: can these vars be removed from this PR?

therealplato:
+1 but see comments @macat

pkg/server/server_test.go
macat: No, actually these are being used.

PR #135, opened 2016-08-24
Remove lock, change delete to stop, defaults

macat:


pkg/instance/instance.go
```diff
@@ -168,7 +168,9 @@ func Unlock(store store.Store, path Path) (*Instance, error) {
 }
 
 func fromJSON(jsonData string) (*Instance, error) {
-	var instance Instance
+	instance := Instance{
+		Lock: false,
```
therealplato: is this a default that gets overwritten by the unmarshal?

pkg/deployment/kubernetes.go
```diff
@@ -53,13 +53,21 @@ type KubernetesDeployment struct {
...
 func NewKubernetesDeployment(config *restclient.Config, playbook *Playbook, variables map[string]string, manifests map[string]*Manifest) (*KubernetesDeployment, error) {
 	var err error
+	vars := map[string]string{}
 	client, err = coreclient.NewForConfig(config)
 	if err != nil {
 		return nil, err
 	}
+	for _, v := range playbook.Vars {
+		var ok bool
+		vars[v], ok = variables[v]
```
therealplato: where do the values of the `variables` map get set?

pkg/instance/instance.go
cored: The zero value of a boolean is false. So you could remove this.

pkg/deployment/kubernetes.go
macat: https://github.com/namely/broadway/blob/master/pkg/services/deployment.go#L67

macat: This will fix the error when the playbook changes and the instances don't have the new variables.

pkg/instance/instance_test.go
```diff
@@ -146,7 +145,11 @@ func TestFindByPlaybookID(t *testing.T) {
 		instances, err := FindByPlaybookPath(tc.Store, tc.PlaybookPath)
 		assert.Equal(t, tc.ExpectedError, err, tc.Scenario)
 		if err == nil {
-			assert.Equal(t, tc.ExpectedInstances, instances, tc.Scenario)
+			expected := []Instance{}
```
therealplato: rename to actual

pkg/deployment/replication_controller.go
```diff
@@ -55,7 +55,7 @@ func deleteRC(namespace, metaName string) error {
 	var i int32
 	rc.Spec.Replicas = &i // Replicas type is *int32 ... so this is *int32(0)
 	client.ReplicationControllers(namespace).Update(rc)
-	time.Sleep(10 * time.Second) // Wait for Kubernetes to delete pods
+	time.Sleep(15 * time.Second) // Wait for Kubernetes to delete pods
```
cored: @macat what about putting this in an environment variable?

poopoothegorilla: will we eventually change this to a watch?

macat: we  need to make this even smarter.

macat: But I'm leaving that for another PR.

pkg/instance/instance.go
```diff
@@ -116,7 +104,7 @@ func AllDeployedAndExpired(store store.Store, path string, expirationDate time.T
 	}
 	for _, i := range instances {
 		if i.Status == StatusDeployed {
-			if i.ExpiredAt <= expirationDate.Unix() {
+			if i.ExpiredAt != 0 && i.ExpiredAt <= expirationDate.Unix() {
```
cored: @macat why do we need to check for `ExpiredAt` been 0 ?

therealplato:
+1

pkg/instance/instance.go
therealplato: because we have old instances that never had this set

macat: @cored If the instance doesn't have this value set, it defaults to 0. So it gets deleted.

pkg/services/slack.go
```diff
@@ -291,26 +253,16 @@ func BuildSlackCommand(cfg cfg.Type, payload string, ds *DeploymentService, is *
...
-	case "lock":
-		if len(terms) < 3 {
-			return &helpCommand{}
-		}
-		return &lockCommand{pID: terms[1], ID: terms[2], Cfg: cfg}
-	case "unlock":
-		if len(terms) < 3 {
-			return &helpCommand{}
-		}
-		return &unlockCommand{pID: terms[1], ID: terms[2], Cfg: cfg}
```
poopoothegorilla: so no more lock feature?

poopoothegorilla:
+1

pkg/instance/instance.go
cored: @therealplato you mean the data that we saved in etcd? for older instances?

cored: @macat if you unmarshall missing fields for your struct the unmarshall will set it to the default value. Could you try without that check ?
```
To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match. Unmarshal will only set exported fields of the struct.
````

pkg/services/slack.go
macat: yup

PR #136, opened 2016-08-25
remove manifest service

macat:


pkg/deployment/manifests.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+)
+
+// LoadManifestFolder returns a map of name:Manifest for each file in root
+func LoadManifestFolder(root, ext string) (map[string]*Manifest, error) {
+	var mm = make(map[string]*Manifest)
+	extRX, err := regexp.Compile(`\.[^.]+$`)
```
poopoothegorilla: do we not care about this error?

macat: yes we do!

poopoothegorilla:
+1 looks good 💃

bobbytables:
:shipit:

PR #137, opened 2016-08-29
Bug fix

orkhanM:
* Once an instance has been deleted status in etcd should be updated
    from "deleting" to "" (or StatusNew)


macat:
We are already setting the status from Slack and Delete API requests delete the instance form ETCD.

PR #138, opened 2016-08-29
Fix missing variables

macat:
My previous change filtered out some basic variables like `id` and `playbook_id`.


poopoothegorilla:
+1

therealplato:
+1

PR #143, opened 2016-11-04
update client kubectl to 1.4.5

ehudc:
@orkhanM suggests this to resolve issues deleting replicationcontrollers when trying to redeploy a broadway bifrost instance


macat:
this is not going to fix it. we need to update the go client.

orkhanM:
@ehudc I meant your local kubectl binary. Like @macat said, the golang client will need to be updated for broadway to be able to delete RCs.

ehudc:
okeydoke then, did the update locally, closing because not relevant

